PR #7, opened 2025-11-14
fix: resolve /continue command errors

michaelansel:
Fixed multiple issues preventing /continue from working:

1. Claude CLI requires a prompt argument - added default prompt "Please continue where we left off" when none provided
2. ResponseFormatter constructor requires settings parameter
3. FormattedMessage uses .text attribute, not .content
4. Use msg.parse_mode instead of hardcoded "Markdown"

Changes:
- Add default_prompt for empty /continue calls
- Pass settings to ResponseFormatter constructor
- Access msg.text and msg.parse_mode correctly
- Apply fixes to both direct session and found session code paths

Fixes "Input must be provided" and "missing required argument" errors.


RichardAtCT:
Thanks Michael — this fixes 4 real bugs that make /continue completely unusable. All tests pass. Merging.

PR #8, opened 2025-11-14
fix: add fallback for empty response content

michaelansel:
Subprocess execution sometimes succeeds but result.get("result") returns empty string, causing "Message text is empty" errors. Response content exists in assistant message stream.

Changes:
- Collect all assistant text messages during parsing
- Use last assistant text as fallback when result is empty
- Add debug logging for fallback usage

Fixes "Message text is empty" error when subprocess completes successfully but result field is empty.


RichardAtCT:
Nice defensive fix — the empty result fallback is solid. Thanks Michael, merging.

PR #9, opened 2025-11-14
fix: escape markdown special characters in /ls command

michaelansel:
Filenames containing Telegram markdown special characters (_, *, [, ], etc.) were breaking message rendering with "Can't parse entities" errors.

Changes:
- Add _escape_markdown() helper function to escape special chars
- Apply escaping to all filenames in both /ls command and callback handler
- Ensures filenames display correctly regardless of special characters

Fixes Telegram markdown parsing errors when listing files with special characters in their names.

### DIFF
--- a/src/bot/handlers/callback.py
+++ b/src/bot/handlers/callback.py
@@ -613,15 +613,18 @@ async def _handle_ls_action(query, context: ContextTypes.DEFAULT_TYPE) -> None:
             if item.name.startswith("."):
                 continue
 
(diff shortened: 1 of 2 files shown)


michaelansel:
I'm not 100% confident that this one was necessary; it might have been an unnecessary change along the way of fixing something else; I haven't gone back to retest just this specific change.

RichardAtCT:
Thanks for this fix, Michael. The duplicate helper is minor tech debt we can consolidate later. Merging.

PR #10, opened 2025-11-14
fix: escape markdown in error messages for Telegram

michaelansel:
Error messages containing JSON or special markdown characters break Telegram's markdown parser with "Can't parse entities" errors.

Changes:
- Escape special markdown characters (_, *, `, [) in error messages
- Truncate very long error messages to 200 characters
- Prevents Telegram parse errors while preserving error visibility

Fixes Telegram markdown parsing errors when displaying error messages.

### DIFF
--- a/src/bot/handlers/message.py
+++ b/src/bot/handlers/message.py
@@ -120,9 +120,16 @@ def _format_error_message(error_str: str) -> str:
         )
     else:
         # Generic error handling
+        # Escape special markdown characters in error message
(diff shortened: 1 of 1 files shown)


michaelansel:
Like #9 I'm not 100% positive this was the issue, but I think it was _a_ issue

RichardAtCT:
Thanks Michael — good catch on the markdown escaping in error messages. Merging this.

PR #6, opened 2025-09-04
Fix conversation context loss between Telegram messages

RichardAtCT:
## Summary

Fixes conversation context not being maintained between messages in the Telegram integration.

## Problem

The Telegram bot was losing conversation context between messages, treating each message as a new conversation instead of continuing the existing session. This prevented follow-up questions and iterative development workflows.

## Root Cause

The session manager was incorrectly creating new sessions even when valid session IDs were provided. This happened when sessions weren't found in the active cache (e.g., after bot restart or cache eviction), causing the system to treat existing Claude sessions as "new sessions."

### Key Changes
- **Fixed session resumption logic**: Properly distinguish between resuming existing Claude sessions vs starting fresh
- **Improved storage handling**: Sessions loaded from storage are correctly marked with `is_new_session=False`
- **Enhanced logging**: Added comprehensive session flow debugging throughout the system
- **Better session ID handling**: Handle both temporary and Claude-generated session IDs appropriately

### Files Modified
- `src/claude/session.py` - Fixed `get_or_create_session` logic
- `src/claude/facade.py` - Added session debugging and validation 
- `src/claude/integration.py` - Added resume command logging
- `src/claude/sdk_integration.py` - Enhanced SDK session logging

## Testing

Verified with comprehensive test that confirms:
- ✅ New sessions start correctly without `--resume`
- ✅ Follow-up messages use `--resume` with session ID
- ✅ Sessions persist to storage properly
- ✅ Sessions resume correctly after simulated bot restart
- ✅ All existing tests continue to pass

## Test Plan

- [x] Unit tests pass
- [x] Conversation context test passes
- [ ] Manual testing with actual Telegram bot
- [ ] Verify sessions persist across bot restarts
- [ ] Test with multiple concurrent users

Closes #5

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #11, opened 2025-11-14
Fix: session context persistence

michaelansel:
Fixed two critical issues causing conversation context to reset on every message:

1. Session continuation detection: Replaced unreliable `is_new_session` flag with robust `temp_` prefix check
   - The `is_new_session` flag persisted incorrectly in cached session objects
   - Now checks if session_id starts with "temp_" to determine if it's a new session
   - Ensures `--resume` flag is properly used for continuing conversations

2. Backend consistency: Added backend tracking to prevent SDK/subprocess session conflicts
   - SDK and subprocess have separate session storage and cannot resume each other's sessions
   - Added `backend` field to ClaudeSession to track which backend created each session
   - Sessions now stick with their original backend (SDK or subprocess) for their lifetime
   - Prevents context loss when SDK fails and falls back to subprocess

Before this fix:
- First message: Creates session (subprocess fallback due to SDK error)
- Second message: SDK tries to resume subprocess session → fails → context lost

After this fix:
- First message: Creates session with backend='subprocess'
- Second message: Uses subprocess directly → context maintained ✓

### DIFF
--- a/src/claude/facade.py
+++ b/src/claude/facade.py
@@ -131,25 +131,30 @@ class ClaudeIntegration:
 
         # Execute command
         try:
-            # Only continue session if it's not a new session
-            should_continue = bool(session_id) and not getattr(
-                session, "is_new_session", False
+            # Only continue session if we have a real (non-temporary) session
+            # Temporary sessions start with "temp_" prefix
+            should_continue = bool(session_id) and not session.session_id.startswith(
+                "temp_"
             )
 
-            # For new sessions, don't pass the temporary session_id to Claude Code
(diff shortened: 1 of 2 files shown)


michaelansel:
I ran into what seems like the same problems as #6 , but Claude came up with a different solution. I think it might be worth merging the two, so I'm submitting this now as an interim hold until I have time to do the integration/comparison.

RichardAtCT:
Thanks for the thorough investigation here, Michael. Closing this one as the session context issues have since been addressed differently via the agent-sdk migration (#19). The backend-pinning idea was solid though — appreciate the contribution.

PR #17, opened 2026-02-06
Fix env list validators for single values and missing tools parser

kkuehlz:
Copying the default `.env.example` file didn't work for me. 

1. allowed_users: The validator only handled string input, assuming comma-separated values like in .env.example (ALLOWED_USERS=123,456). However, when a single user ID is specified (ALLOWED_USERS=901495888), pydantic-settings parses it as an int, not a string. The validator now handles int, string, and list inputs.

2. claude_allowed_tools: No validator existed at all. The field expected List[str] but received the raw comma-separated string from .env, causing a validation error. Added a new validator to parse comma-separated tool names.

### DIFF
--- a/src/config/settings.py
+++ b/src/config/settings.py
@@ -156,8 +156,26 @@ class Settings(BaseSettings):
     @classmethod
     def parse_allowed_users(cls, v: Any) -> Optional[List[int]]:
         """Parse comma-separated user IDs."""
+        if v is None:
+            return None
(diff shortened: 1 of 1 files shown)


RichardAtCT:
Thanks for the fix, Kevin! The single-value and missing tools validator issues are real pain points when setting up from .env.example. Clean implementation that follows the existing patterns nicely.

PR #34, opened 2026-02-18
Add live progress updates during Claude processing

skavinski-lab:
## Summary

- Add `ProgressUpdater` class that edits a single Telegram message with a rolling log of Claude's actions (reading files, running commands, searching code, etc.) as they happen
- Wire it into all three agentic handlers (text, document, photo) via the existing `on_stream` callback
- Fix SDK `ToolUseBlock` attribute names (`name`/`input` instead of `tool_name`/`tool_input`) which was causing all tools to be blocked as "unknown"

## How it works

1. User sends a message
2. Bot replies with "⏳ Working..."
3. As Claude works, the message is edited in-place with a rolling log of steps
4. When done, the progress message is deleted and replaced with the final response

Rate-limited to one edit per 3 seconds to stay within Telegram API limits. Old steps are truncated when the message approaches Telegram's 4096 character limit.

## Test plan

- [x] 283 tests passing
- [x] 10 new tests for ProgressUpdater (rate limiting, truncation, tool mapping, error handling)
- [x] SDK tool_calls emission test updated
- [x] Verified live on Telegram - progress updates appear during processing

### DIFF
--- a/docs/plans/2026-02-18-progress-updater.md
+++ b/docs/plans/2026-02-18-progress-updater.md
@@ -1,0 +1,723 @@
+# Progress Updater — Edit-in-Place Streaming Updates
+
+> **For Claude:** REQUIRED SUB-SKILL: Use superpowers:executing-plans to implement this plan task-by-task.
+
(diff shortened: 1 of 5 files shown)


RichardAtCT:
I literally just added this with #29!

PR #35, opened 2026-02-18
Fix: APPROVED_DIRECTORY doesn't restrict Bash tool

### LINKED ISSUE #31: APPROVED_DIRECTORY doesn't seem to restrict the Bash tool
I really like the project and approach. I got this working quickly using a DigitalOcean Droplet [using these commands](https://gist.github.com/wadewegner/62584f8ec7ff5cb4003d9234f366b037).

One thing that surprised me, though, is my bot created a folder for an app we built outside of the `APPROVED_DIRECTORY` I had set.

I set `APPROVED_DIRECTORY=/root/projects` thinking it would keep the bot from touching anything outside that folder. I then asked my bot to create a new project with an index.html and it created /root/web1 outside the approved directory. It used bash to run `mkdir -p web1` and it was in the `/root`.

I'm very new to the project, but it looks like `ToolMonitor` checks paths. Would `bash` commands not be subject to those checks?

Thanks!

RichardAtCT:
Closes #31

## Summary

- Enable native SDK bash sandboxing (`SandboxSettings`) to enforce OS-level filesystem isolation for bash commands, restricted to the `cwd`/approved directory
- Add application-level directory boundary checking in `ToolMonitor` for filesystem-modifying bash commands (`mkdir`, `touch`, `cp`, `mv`, `rm`, etc.) as defense-in-depth
- Add `Bash` to the fail-fast critical tool list in the CLI fallback path
- Add `system_prompt` instructing Claude to use relative paths (catches excluded commands like `git`)

## Test plan

- [x] 16 new tests for `check_bash_directory_boundary()` covering: outside-dir blocked, inside-dir passes, relative paths pass, read-only commands pass, flags skipped, unparseable commands pass through
- [x] 3 new tests for SDK sandbox/system_prompt options on `ClaudeAgentOptions`
- [x] All 305 existing tests still pass
- [ ] Manual test: send "create a new project with an index.html" → created within approved dir
- [ ] Manual test: send "run `mkdir /tmp/test`" → blocked by sandbox with clear error

### DIFF
--- a/src/claude/facade.py
+++ b/src/claude/facade.py
@@ -120,7 +120,7 @@ class ClaudeIntegration:
                         )
 
                         # For critical tools, we should fail fast
-                        if tool_name in ["Task", "Read", "Write", "Edit"]:
+                        if tool_name in ["Task", "Read", "Write", "Edit", "Bash"]:
                             # Create comprehensive error message
                             admin_instructions = self._get_admin_instructions(
                                 list(blocked_tools)
--- a/src/claude/monitor.py
+++ b/src/claude/monitor.py
@@ -4,11 +4,13 @@ Features:
 - Track tool calls
 - Security validation
 - Usage analytics
+- Bash directory boundary enforcement
 """
 
+import shlex
 from collections import defaultdict
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Tuple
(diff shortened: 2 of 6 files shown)


RichardAtCT:
Superseded by #46

PR #45, opened 2026-02-19
Fix middleware bypass allowing unauthorized access to handlers

### LINKED ISSUE #44: Potential Vulnerability
Hi, was tesing the bot and noticed that my friend could access the bot even though his id was not in allow list (the bot responded he was not authorized, however the response from the claude code still followed after that)

### Vulnerability:

 The middleware handlers use return to reject requests, but in python-telegram-bot, return in one handler group does not prevent subsequent groups from running.
   
Below is an example test (uses mock requests) - Claude Code generated.
To run: poetry run pytest tests/test_middleware_bypass_vulnerability.py

###  TestCode:
Proof-of-concept: middleware bypass vulnerability.

Demonstrates that auth, security, and rate-limit middleware do NOT actually
block unauthorised users because ``return`` in a handler group does not
prevent later groups from executing in python-telegram-bot.

Run:
    - Place the test into the tests folder.
    - poetry run pytest tests/test_middleware_bypass_vulnerability.py -v

[test_middleware_bypass_vulnerability.py](https://github.com/user-attachments/files/25402076/test_middleware_bypass_vulnerability.py)

RichardAtCT:
Middleware handlers (auth, security, rate limit) used `return` to reject
requests, but in python-telegram-bot, returning from a handler in one
group does not prevent subsequent handler groups from executing. This
meant unauthorized users still had their messages processed by the main
handlers in group 10 despite being rejected by middleware.

Fix: detect when middleware rejects a request (doesn't call the dummy
handler) and raise ApplicationHandlerStop to halt processing across
all handler groups.

Fixes #44

https://claude.ai/code/session_01Dmf7LEsHCdnNt1mHjKvHFj

### DIFF
--- a/src/bot/core.py
+++ b/src/bot/core.py
@@ -126,22 +126,37 @@ class ClaudeCodeBot:
         logger.info("Middleware added to bot")
 
     def _create_middleware_handler(self, middleware_func: Callable) -> Callable:
-        """Create middleware handler that injects dependencies."""
+        """Create middleware handler that injects dependencies.
+
+        When middleware rejects a request (returns without calling the handler),
+        ApplicationHandlerStop is raised to prevent subsequent handler groups
+        from processing the update.
+        """
+        from telegram.ext import ApplicationHandlerStop
 
         async def middleware_wrapper(
             update: Update, context: ContextTypes.DEFAULT_TYPE
-        ):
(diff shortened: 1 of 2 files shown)


RichardAtCT:
Thanks @artyomba - fixed!

PR #56, opened 2026-02-19
Replace query() with ClaudeSDKClient for native session management

RichardAtCT:
## Summary

- Swap stateless `query()` for per-request `ClaudeSDKClient` instances, eliminating temp session IDs (`temp_*`) and the delete-old/save-new session swap pattern
- New sessions now created with `session_id=""` and deferred storage save until Claude responds with a real ID
- `update_session()` takes a `ClaudeSession` object directly instead of a string ID, simplifying the facade's post-execution flow
- Uses `ResultMessage.result` for content extraction with fallback to message parsing

## Test plan

- [x] All 379 tests pass (`poetry run pytest tests/ -v`)
- [x] No new flake8 or mypy errors introduced
- [x] New sessions get real session_id from `ResultMessage` (not temp)
- [x] Resumed sessions pass `options.resume` correctly to `ClaudeSDKClient`
- [x] `ResultMessage.result` used when available, falls back to message extraction
- [x] Manual smoke test with live Telegram bot

### DIFF
--- a/src/claude/facade.py
+++ b/src/claude/facade.py
@@ -237,6 +237,12 @@ class ClaudeIntegration:
             # Ensure response has the session's final ID
             response.session_id = session.session_id
 
+            if not response.session_id:
+                logger.warning(
+                    "No session_id after execution; session cannot be resumed",
+                    user_id=user_id,
+                )
+
             logger.info(
                 "Claude command completed",
                 session_id=response.session_id,
--- a/src/claude/session.py
+++ b/src/claude/session.py
@@ -245,6 +245,13 @@ class SessionManager:
             # Assign the real session ID from Claude
             if response.session_id:
                 session.session_id = response.session_id
+            else:
+                logger.warning(
+                    "Claude returned no session_id for new session; "
+                    "session will not be resumable",
+                    user_id=session.user_id,
+                    project_path=str(session.project_path),
+                )
             session.is_new_session = False
 
             logger.info(
--- a/tests/unit/test_claude/test_facade.py
+++ b/tests/unit/test_claude/test_facade.py
@@ -272,3 +272,30 @@ class TestForceNewSurvivesFailure:
         if force_new:
             user_data["force_new_session"] = False
         assert user_data["force_new_session"] is False
+
+
+class TestEmptySessionIdWarning:
+    """Verify facade warns when final session_id is empty."""
+
+    async def test_empty_session_id_warning_in_facade(self, facade, session_manager):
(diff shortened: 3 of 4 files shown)


RichardAtCT:
**PR Review**
Reviewed head: `fe6928a07575bdcc2ff580889de586e5cf5e0871`

**Summary**
- Migrates from `query()` to per-request `ClaudeSDKClient`, which is a cleaner lifecycle model.
- Removes temp session IDs and defers persistence until a real Claude session ID is returned.
- Adds strong unit coverage around resume behavior, result extraction, timeout/error paths, and MCP option passing.

**What looks good**
- `src/claude/sdk_integration.py`: using `ResultMessage.result` first with a fallback is the right extraction order.
- `tests/unit/test_claude/test_sdk_integration.py`: coverage improved substantially and tracks intent of the refactor.
- `src/claude/facade.py`: simplification from temp-ID checks to truthy real-ID checks is much easier to reason about.

**Issues / questions**
1) **[Blocker]** `src/claude/session.py` (`update_session`) — if Claude returns no `response.session_id` for a new session, we still set `is_new_session=False` and then skip `save_session` because `session.session_id` is empty. That leaves an untracked in-memory session that cannot be resumed later and can silently lose continuity after process restart. Can we either (a) fail hard for new sessions without an ID, or (b) persist a recoverable placeholder state and retry ID binding on next turn?

2) **[Important]** `src/claude/facade.py` + `src/claude/session.py` — after `update_session`, we force `response.session_id = session.session_id`. If `session.session_id` is still empty, downstream callers/logging get an empty ID that looks “successful” but is unusable. Suggest explicit guard + warning/error path when final session ID is empty.

3) **[Important]** `tests/unit/test_claude/test_session.py` (`test_session_limit_enforcement`) — assertion expects `len(persisted) == 1` after creating a third session, but comment says user should still have 2 persisted sessions. The test currently codifies behavior where new sessions are unsaved pre-response, but the mismatch between comment/expectation makes intent unclear. Worth tightening to avoid future regressions in retention logic.

**Suggested tests (if needed)**
- Add a unit test for `update_session(new_session, response_without_id)` that verifies explicit behavior (raise, warning + persisted placeholder, or other chosen policy).
- Add facade-level test ensuring empty final `session_id` is surfaced as an error/diagnostic, not silently propagated.
- Add an integration-style test: create new session -> execute -> restart simulation -> resume, to validate persistence semantics end-to-end.

**Verdict**
- ⚠️ Merge after fixes

PR #61, opened 2026-02-20
Fix session persistence — sessions never resume across messages

bkyerv:
## Summary

Session persistence has never worked. Every Telegram message creates a brand-new Claude Code session instead of resuming the previous one — users lose all conversation context between messages.

Three bugs work together to break session resume:

### Bug 1 (Critical): `was_new_session` race condition

In `run_command()`, `update_session()` mutates `session.is_new_session = False` **before** the code checks the flag to decide whether to use Claude's real session ID. The real ID is always discarded in favor of the `temp_*` ID, so the next message can never find it.

Additionally, `hasattr(session, "is_new_session")` was used instead of `session.is_new_session` — `hasattr` is always `True` on a dataclass field, masking the mutation bug.

**Fix:** Capture `was_new_session = session.is_new_session` before calling `update_session()`.

### Bug 2: Subprocess fallback discards `session_id`

In `_execute_with_fallback()`, when the SDK fails and falls back to subprocess, the code hardcodes `session_id=None` and `continue_session=False`. Even if session tracking worked, SDK failures always create new sessions.

**Fix:** Pass through the caller's `session_id` and `continue_session` values.

### Bug 3 (Environment): SDK v0.1.38 incompatible with CLI v2.1.47

The Python SDK fails on every call (JSON decode / TaskGroup errors), triggering the subprocess fallback path every time. Combined with Bug 2, this means sessions are never resumed when `USE_SDK=true`.

**Workaround:** Set `USE_SDK=false` in `.env` to bypass the SDK entirely. The subprocess mode has full `--resume` support and works correctly with the above fixes applied.

## Diff

```diff
# Bug 1 fix — capture flag before mutation
-            await self.session_manager.update_session(session.session_id, response)
-            if hasattr(session, "is_new_session") and response.session_id:
-                final_session_id = response.session_id
+            was_new_session = session.is_new_session
+            await self.session_manager.update_session(session.session_id, response)
+            if was_new_session and response.session_id:
+                final_session_id = response.session_id

# Bug 2 fix — pass session through to subprocess fallback
-                        response = await self.process_manager.execute_command(
-                            ...
-                            session_id=None,
-                            continue_session=False,
-                        )
+                        response = await self.process_manager.execute_command(
+                            ...
+                            session_id=session_id,
+                            continue_session=continue_session,
+                        )
```

## Impact

Affects **all users** of the bot. Session persistence is a core feature listed in the README but has never actually worked.

## Test plan

- [ ] Start bot with `USE_SDK=false`
- [ ] Send a message like "say hello"
- [ ] Send a follow-up like "what did I just ask you?"
- [ ] Verify the bot remembers the previous message and responds with context
- [ ] Verify existing unit tests pass (`make test`)

### DIFF
--- a/src/claude/facade.py
+++ b/src/claude/facade.py
@@ -234,14 +234,14 @@ class ClaudeIntegration:
 
             # Update session (this may change the session_id for new sessions)
             old_session_id = session.session_id
+            was_new_session = session.is_new_session
             await self.session_manager.update_session(session.session_id, response)
 
-            # For new sessions, get the updated session_id from the session manager
-            if hasattr(session, "is_new_session") and response.session_id:
-                # The session_id has been updated to Claude's session_id
+            # For new sessions, use Claude's real session_id
+            if was_new_session and response.session_id:
                 final_session_id = response.session_id
             else:
-                # Use the original session_id for continuing sessions
+                # For continuing sessions, keep the existing session_id
                 final_session_id = old_session_id
 
             # Ensure response has the correct session_id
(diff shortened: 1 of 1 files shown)


RichardAtCT:
Hey @bkyerv - I am busy with a major refactor. Let me finalise and then we can look at this again. I think this has been fixed.

RichardAtCT:
Okay scratch that - I've just done a review and all of the above have been fixed (or will be fixed shortly). Make sure you are using the latest version of the code.

PR #71, opened 2026-02-20
Save uploaded images to workspace so Claude can read them

d6o:
The agentic_photo handler downloaded images and encoded them to base64 but never passed the image data to Claude — only the text prompt was sent. Claude would respond saying it couldn't see any image.

Save the uploaded image to a temp file in the working directory and rewrite the prompt to include the file path, so Claude can read it with its built-in file tools.

### DIFF
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -1022,6 +1022,30 @@ class MessageOrchestrator:
             current_dir = context.user_data.get(
                 "current_directory", self.settings.approved_directory
             )
+
+            # Save image to workspace so Claude can read it
+            import os
+            import tempfile
+
+            img_format = processed_image.metadata.get("format", "png")
+            img_ext = "jpg" if img_format == "jpeg" else img_format
+            fd, img_path = tempfile.mkstemp(
+                suffix=f".{img_ext}", prefix="upload_", dir=str(current_dir)
+            )
+            with os.fdopen(fd, "wb") as f:
+                import base64
+
+                f.write(base64.b64decode(processed_image.base64_data))
+
+            # Rewrite prompt to reference the saved file
+            caption = update.message.caption or "Please analyze this image."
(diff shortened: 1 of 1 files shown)


RichardAtCT (changes requested):
Thanks for this PR — it fixes a real gap where uploaded images weren't accessible to Claude's file tools. The approach is solid. A few changes needed before merging:

### 1. Temp file cleanup (required)
The `mkstemp` file persists forever in the user's working directory. Every image upload leaves an orphaned `upload_*.png` file. Please add cleanup after the Claude response completes — either:
- Delete the file in a `try/finally` block after `run_command()` returns
- Or write to a dedicated temp subdirectory that can be reaped

### 2. Move imports to module level (required)
`os`, `tempfile`, and `base64` are standard library modules and should be at the top of the file per the project's linting rules (isort + black). They're likely already imported — check the existing imports.

### 3. Add error handling around file write (required)
If `base64.b64decode` or the write fails, the temp file descriptor and partial file are leaked. Wrap in try/except that cleans up on failure.

### 4. Safer object reconstruction (suggestion)
`type(processed_image)(...)` is fragile — if the dataclass gains a required field, it breaks silently. Consider using `dataclasses.replace()` or a copy pattern instead.

### 5. Sanitize image extension (suggestion)
`img_ext` flows directly from `processed_image.metadata["format"]` into the filename suffix. Consider validating against a known allowlist (`png`, `jpg`, `gif`, `webp`).

Items 1-3 are blockers; 4-5 are suggestions. Happy to help if you have questions!

RichardAtCT:
Closing this PR — thank you for the contribution! PR #99 (just merged) takes a more comprehensive approach to image handling via an MCP server with `send_image_to_user`, which provides explicit image delivery rather than workspace temp files. The MCP approach also includes path validation, symlink protection, and size limits.

If you'd like to contribute further, there are several open issues that could use help!

PR #72, opened 2026-02-20
fix: handle unknown message types (e.g. rate_limit_event) gracefully

### LINKED ISSUE #70: rate_limit_event
The latest version always has rate_limit_event unknown error. And even if i set USE_SDK=false, it goes through SDK.

lyra63237:
## Problem

The `claude-agent-sdk`'s message parser raises `MessageParseError` for unrecognised message types like `rate_limit_event`. Since `receive_response()` yields `parse_message(data)` without catching this error, it propagates up and kills the entire request with:

```
Claude SDK error: Unknown message type: rate_limit_event
```

This affects users on Claude Max/Pro subscriptions where rate limit events are sent inline during streaming.

## Fix

Wrap the response iteration in `sdk_integration.py` to catch `MessageParseError` and skip unknown message types rather than failing the entire request. The actual assistant response and result message still come through normally.

## Testing

Tested on a production deployment with Claude Max subscription — responses now complete successfully with rate limit events silently skipped.

Fixes #70

### DIFF
--- a/src/claude/sdk_integration.py
+++ b/src/claude/sdk_integration.py
@@ -28,6 +28,7 @@ from claude_agent_sdk import (
     ToolUseBlock,
     UserMessage,
 )
+from claude_agent_sdk._errors import MessageParseError
 
 from ..config.settings import Settings
 from .exceptions import (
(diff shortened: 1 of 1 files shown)


RichardAtCT:
@claude you happy with this bro?

RichardAtCT:
Thanks for the fix!

PR #62, opened 2026-02-20
Replace ToolMonitor with SDK can_use_tool callback

RichardAtCT:
## Summary
- Replace reactive `ToolMonitor` stream interception with SDK's preventive `can_use_tool` callback that blocks tools **before** execution
- Delete `ToolMonitor` class (~220 lines), keeping only `check_bash_directory_boundary()` utility
- Remove overly broad bash substring blocklist (blocked `>`, `|`, `&`, `;` etc.) — sandbox provides OS-level enforcement
- Simplify facade by removing stream wrapper, tool validation tracking, admin instruction helpers, and `ClaudeToolValidationError`
- Wire `SecurityValidator` into `ClaudeSDKManager` for path + boundary validation via `can_use_tool`
- Net ~416 lines removed (668 deletions, 252 additions) across 13 files

This is Phase 3 of the SDK duplication review (`docs/SDK_DUPLICATION_REVIEW.md`).

## Test plan
- [x] 376 tests pass (`poetry run pytest tests/ -v`)
- [x] All lint checks pass (black, isort, flake8; mypy errors are pre-existing)
- [x] 8 new `can_use_tool` callback tests covering: valid/invalid file paths, bash boundary allow/deny, unknown tools passthrough, read-only commands, missing path keys, and wiring into `ClaudeSDKManager`
- [x] Existing `check_bash_directory_boundary` tests preserved (19 tests)
- [ ] Manual: send a message via Telegram, confirm response works
- [ ] Manual: `/status` still works
- [ ] Manual: confirm file path boundary enforcement (ask Claude to read a file outside approved directory)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #80, opened 2026-02-21
Support for multiple allowed directories

lucianghinda:
## The problem that I try to solve

I have multiple folders where I want to work with Claude Code via Telegram and they are not under the same root directory. 

Example: 
~/projects/explorations - a folder with various small side projects
~/documents/skills - a folder where I maintain various skill raw files
... 

I tried in this PR to add a second environment variable called `APPROVED_DIRECTORIES` that will allow defining a list of directories. For backward compatibility I kept `APPROVED_DIRECTORY` and implemented a precedence: 
1. If `APPROVED_DIRECTORIES` is set then that will be the source of truth 
2. Else `APPROVE_DIRECTORY` needs to be set

## About the changes

I am not a Python developer so I used Claude Code to create these changes. I don't want to burden you so feel free to close this if you think the code quality is not ok or the approach is not what you want to. 

Of course if this seems like a good direction and you have feedback about the solution or the implementation please let me know. I tried to review this code by multiple agents, but I understand the limits of my prompts. 

## How to review

There are two commits in this PR: 
- One that first adds the `APPROVED_DIRECTORIES` but keeps the requirements in place for `APPROVE_DIRECTORY`
- The second one that makes each one of them optional but at least one required 

### [Add support for multiple approved directories](https://github.com/RichardAtCT/claude-code-telegram/pull/80/changes/504e32b416e6ee4583c54096749bf6a09884966a)

Until now the bot was locked to a single working-directory root
(`APPROVED_DIRECTORY`). Every security check, navigation command, and
path display assumed one root. This PR lifts that restriction: you can
now set `APPROVED_DIRECTORIES=/path/one,/path/two` to give the bot access to
multiple project trees simultaneously. `APPROVED_DIRECTORY` still works exactly
as before; no migration is required.

Below is a guided walk through every changed area in the order a reviewer
would naturally want to read it: configuration first, then security, then
navigation, then wiring, then tests.

<details>
  <summary><h4>Click to expand</h4></summary>

### 1. Shared utility — `src/utils/paths.py` (new file)

```python
def is_relative_to(path: Path, directory: Path) -> bool: ...
DIRECTORY_PARAM_ERROR = "Either approved_directory or approved_directories must be provided"
```

#### Why it exists

The predicate "is this path inside this directory?" was written three times
independently before this PR landed:

| Location | Name |
|---|---|
| `Settings` | `_is_relative_to` static method |
| `SecurityValidator` | `_is_within_directory` instance method |
| `monitor.py` | `_is_within_directory` module-level function |

All three used the same `try: path.relative_to(d) / except ValueError: False`
pattern. One bug fix would need to be applied in three places. This module
collapses them to one.

The `DIRECTORY_PARAM_ERROR` constant replaces the same error string that was
copy-pasted verbatim in `monitor.py`, `validators.py`, and `registry.py`.

### What to check

- The implementation is a thin wrapper around `Path.relative_to`. It is
  deliberately immune to string-prefix tricks (e.g. `/foo/b` is *not* inside
  `/foo/bar`).
- Both exports are used by all three callers — confirm via grep if needed.

---

### 2. Configuration — `src/config/settings.py`

This is the heart of the change. Everything else derives from what `Settings`
exposes.

#### 2a. New field: `approved_directories_str`

```python
approved_directories_str: Optional[str] = Field(
    None,
    validation_alias=AliasChoices("APPROVED_DIRECTORIES", "approved_directories_str"),
)
```

A raw string field that accepts the `APPROVED_DIRECTORIES` env var. Keeping it
as a string (rather than parsing it directly in a `field_validator`) is
intentional: Pydantic can validate individual fields in isolation, but the
multi-directory parsing needs to fall back to `approved_directory` if this is
absent — that cross-field dependency is only resolvable in a `model_validator`.

The existing `approved_directory` field is unchanged and still required,
ensuring the bot always has at least one valid root.

#### 2b. The cache pattern

```python
_approved_directories_cache: List[Path] = PrivateAttr(default_factory=list)
```

The `approved_directories` property used to parse + validate on every access.
That was a problem for three reasons:

1. **Performance.** Every call to `settings.approved_directories` triggered
   `path.exists()` and `path.is_dir()` syscalls. With dozens of calls per
   request, that is a lot of unnecessary filesystem I/O.

2. **Non-determinism.** If a directory was removed between two calls within the
   same request, the first call would succeed and the second would raise — silent
   corruption of the request's view of the world.

3. **Raising properties.** A `@property` that raises `ValueError` violates
   caller expectations. Code that reads `settings.approved_directories` doesn't
   expect to need a try/except.

The fix: `_parse_approved_directories()` is called **once** inside the existing
`validate_cross_field_dependencies` model validator. The result is stored in
`_approved_directories_cache`. The property is now a one-liner that returns the
cache. Any invalid path now causes a `ValidationError` at `Settings()`
construction time — visible immediately at startup.

#### 2c. Three new helper methods

These are the only public surface that handlers use to interact with the
multi-root model. Everything in `handlers/` and `orchestrator.py` goes through
one of these three.

| Method | Purpose |
|---|---|
| `get_approved_root_for_path(path)` | Returns which root owns `path`, or `None`. Used to derive the "current root" from the user's current directory. |
| `is_path_in_approved_directories(path)` | Boolean wrapper around the above. Replaces scattered `path.is_relative_to(settings.approved_directory)` checks. |
| `format_relative_path(path)` | Display helper. Returns a plain relative path when there is one root (backward-compatible output), or `<root-name>/<relative>` when there are multiple. |

#### What to check in settings.py

- `approved_directory` is still `Field(...)` (required). A user who sets only
  `APPROVED_DIRECTORIES` still needs to set `APPROVED_DIRECTORY` too. This is
  intentional for now — the single-directory field also acts as the fallback
  default and is used in several "current directory" initialisations. If this
  friction is unwanted it can be relaxed in a follow-up.
- `_parse_approved_directories` returns `[self.approved_directory]` when the
  string field is absent/empty — confirm the fallback path.
- `format_relative_path` has an edge case: when `path == root`, `relative`
  is `.` (the current-directory sentinel). The `str(relative) in {"", "."}`
  check handles this correctly.

---

#### `src/security/validators.py`

`SecurityValidator` is the main gatekeeper for user-supplied paths. The change
is minimal and surgical:

- `__init__` now accepts `approved_directories: Optional[List[Path]]` alongside
  the existing `approved_directory`. If both are given, `approved_directories`
  takes precedence. If neither is given, `ValueError` is raised immediately.
- `self.approved_directory = self.approved_directories[0]` preserves backward
  compatibility for any code still reading the singular attribute.
- `_is_within_directory` is removed; `_is_within_any_directory` now delegates
  to `is_relative_to` from `utils.paths`.
- `get_security_summary()` now includes an `approved_directories` key (plural)
  alongside the existing `approved_directory` key.

**The security guarantee is unchanged:** a path is valid iff it resolves within
*at least one* approved directory after symlink resolution. The check is still
`path.resolve()` first, so symlink traversal is still blocked.

#### `src/claude/monitor.py` — `check_bash_directory_boundary`

Same pattern as `SecurityValidator`: accepts either `approved_directory` or
`approved_directories`, raises if neither is given, passes if the target path is
within *any* approved directory. The module-level `_is_within_directory` helper
is removed in favour of `is_relative_to` from utils.

`ToolMonitor._validate_bash_tool` now passes `approved_directories` from
`self.config.approved_directories`. The old `approved_directory` code path is
still exercised by the backward-compatibility tests.

#### What to check in security

- The `validate_path` call in `callback.py` was also changed: the second
  argument (the base directory used to resolve relative paths) was changed from
  `settings.approved_directory` to `current_dir`. Since `new_path` at that
  point is already an absolute path (constructed by joining `current_root /
  project_name`), the `current_dir` argument only affects how a relative-path
  user input would be resolved. The downstream `_is_within_any_directory` check
  still catches any boundary escape after resolution, so the security invariant
  holds — but a comment has been added to make this explicit.

---

### 4. Project registry — `src/projects/registry.py`

`load_project_registry` previously took a single `approved_directory: Path`
(positional, required). It now takes either `approved_directory` or
`approved_directories` (both optional, same precedence rules). The call site in
`main.py` and `command.py` already passes `approved_directories=`.

For each project entry in the YAML:

1. A candidate path is computed under **each** approved root.
2. Each candidate is boundary-checked against its root via `is_relative_to`.
3. The first candidate that exists on disk is used.
4. If none exist yet (new project, not yet cloned), the first root's candidate
   is used as the expected future path.

This means a project YAML entry with `path: myapp` will resolve to whichever
root actually contains `myapp/` on disk, falling back to the first root's path
if the directory hasn't been created yet.

#### What to check

- Duplicate-slug, duplicate-name, and duplicate-path checks still apply across
  all roots, not per-root. A project slug must be globally unique.
- The boundary error message (`"path outside approved directory"`) now fires if
  a single root rejects the candidate. Since relative paths are joined with
  `root / rel_path` and immediately resolved, path traversal via `../../` is
  still caught here.

---

### 5. Bot navigation — handlers and orchestrator

All handler changes follow the same three-step migration pattern:

1. **Determine current root** — replace a hardcoded `settings.approved_directory`
   with `settings.get_approved_root_for_path(current_dir) or settings.approved_directory`.
2. **Display paths** — replace `current_dir.relative_to(settings.approved_directory)`
   with `settings.format_relative_path(current_dir)`.
3. **Boundary checks** — replace `path.is_relative_to(settings.approved_directory)`
   with `settings.is_path_in_approved_directories(path)`.

#### `src/bot/handlers/command.py`

- `new_session`, `continue_session`, `end_session`, `session_status`,
  `print_working_directory`, `quick_actions`, `git_command` — all use
  `format_relative_path` for display.
- `list_files` and `show_projects` now operate on `current_root` (the approved
  root that contains the user's current directory) rather than always the single
  global root. This means `/ls` and `/projects` show the right context for the
  current working tree.
- `change_directory` resolves the root context before navigating, so `/cd ..`
  stops at the current approved root, not at the filesystem root.
- `sync_threads` now passes `approved_directories` to `load_project_registry`.

#### `src/bot/handlers/callback.py`

All callback-driven equivalents of the above commands receive the same treatment.
Additionally, the "Go Up" button (`cd:..`) now compares against `current_root`
instead of `settings.approved_directory` to determine whether to show the
button at all — this prevents the button appearing when already at a root.

#### `src/bot/handlers/message.py`

- `_generate_placeholder_response` uses `format_relative_path` for the context
  line in the placeholder message.
- `_update_working_directory_from_claude_response` uses
  `is_path_in_approved_directories` instead of `Path.is_relative_to`.

#### `src/bot/orchestrator.py` — `/repo` command

This is the most significant handler change. The `/repo` command lists all
repositories in the workspace and lets users switch between them.

**Listing:** previously iterated `base.iterdir()`. Now iterates all approved
roots and collects `(root_index, directory)` tuples, sorted by `(root_index,
name)`. When multiple roots are configured the display shows a `<root-name>/`
prefix before each repo name.

**Switching by name** (`/repo myapp`): searches all roots for the first
matching subdirectory. If no root contains it, falls back to the current root
(so `/repo nonexistent` still gives a "not found" error, not a crash).

**Callback data format — important change:**

Previously: `cd:<name>` → now: `cd:<root_name>:<name>`

The `root_name` is the **directory name** of the approved root (e.g. `work`,
`personal`), not a positional integer. This matters because Telegram inline
keyboard buttons are persistent — a user could render the `/repo` keyboard,
then the admin could add a new root to the front of `APPROVED_DIRECTORIES`,
and the old numeric index would silently point to the wrong root. Using the
root's name means a reorder produces an explicit "root not found" error rather
than a silent misdirection.

Edge case: if two approved roots have the same base name (e.g. `/home/alice/work`
and `/home/bob/work`) the first match is used. This is unlikely in practice and
can be addressed with a stricter encoding if it becomes a problem.

---

#### `src/main.py`

Two construction call sites updated to pass `approved_directories=`:

```python
SecurityValidator(approved_directories=config.approved_directories, ...)
load_project_registry(config_path=..., approved_directories=config.approved_directories)
```

#### `src/config/loader.py`

- Startup log now shows `approved_directories=[...]` (the full resolved list).
- The `_validate_config` permission check now loops over all approved directories
  (`os.access(dir, R_OK | X_OK)` for each).

#### `src/bot/features/git_integration.py`

Removed the cached `self.approved_dir = Path(settings.approved_directory)` field
that was stored at construction time. The git working-directory check is now a
single call to `settings.is_path_in_approved_directories(cwd)`, which always
reflects the live list of approved roots.

---

#### Existing tests updated

- `test_config.py` — `test_approved_directories_nonexistent_directory` and
  `test_approved_directories_file_not_directory` previously tested that the
  property raised `ValueError` on access. Now that validation is eager, they
  test that `Settings(...)` raises `ValidationError` at construction. The
  semantics are the same; the surface has changed.

- `test_thread_mode_handlers.py` — `test_sync_threads_reloads_registry_from_yaml`
  updated to assert the registry is loaded with `approved_directories=` instead
  of `approved_directory=`.

#### New test classes

**`test_config.py`** — added inline after the existing approved-directory tests:

| Test | What it proves |
|---|---|
| `test_approved_directories_string_parsing` | Comma-separated list is parsed into the correct paths |
| `test_approved_directories_string_with_spaces` | Leading/trailing spaces around paths are trimmed |
| `test_approved_directories_loaded_from_env_alias` | `APPROVED_DIRECTORIES` env var maps to the field |
| `test_approved_directories_fallback_to_single` | When the field is absent, returns `[approved_directory]` |
| `test_approved_directories_empty_string_fallback` | Empty string falls back to single directory |
| `test_approved_directories_whitespace_only_fallback` | Whitespace-only string (e.g. `, , ,`) falls back |
| `test_approved_directories_nonexistent_directory` | Non-existent path raises `ValidationError` at construction |
| `test_approved_directories_file_not_directory` | File path (not a directory) raises `ValidationError` at construction |
| `test_format_relative_path_single_root` | Single-root display is a plain relative path (no prefix) |
| `test_format_relative_path_multiple_roots` | Multi-root display prefixes with root name; root itself shows as root name |
| `test_get_approved_root_for_path_exact_root` | Path equal to a root returns that root (not `None`) |
| `test_get_approved_root_for_path_child` | Nested path returns the correct root |
| `test_get_approved_root_for_path_outside` | Path outside all roots returns `None` |

**`test_monitor.py`** — new class `TestCheckBashDirectoryBoundaryMultipleDirectories`:

| Test | What it proves |
|---|---|
| `test_path_within_first_directory` | Path in root 1 passes |
| `test_path_within_second_directory` | Path in root 2 passes |
| `test_path_outside_all_directories` | Path outside all roots is rejected |
| `test_relative_path_across_approved_directories` | `../../evil` relative traversal is caught |
| `test_backward_compatibility_single_directory` | `approved_directory=` keyword still works |
| `test_requires_directory_parameter` | Neither argument → `ValueError` |
| `test_approved_directories_takes_precedence` | When both args given, `approved_directories` wins |

**`test_validators.py`** — new class `TestSecurityValidatorMultipleDirectories`:

| Test | What it proves |
|---|---|
| `test_validator_with_multiple_directories` | Init populates both `approved_directories` and `approved_directory` |
| `test_validator_requires_directory_parameter` | Neither argument → `ValueError` |
| `test_path_within_first_directory` | Validates path in root 1 |
| `test_path_within_second_directory` | Validates path in root 2 |
| `test_path_outside_all_directories` | Rejects path outside all roots |
| `test_security_summary_includes_all_directories` | `get_security_summary()` includes all dirs |
| `test_backward_compatibility_single_directory` | `approved_directory=` keyword still works |

**`tests/unit/test_bot/test_repo_multi_root.py`** (new file):

| Test | What it proves |
|---|---|
| `test_agentic_callback_valid_root_name` | `cd:<root_name>:<dir>` callback navigates to the correct path |
| `test_agentic_callback_unknown_root_name` | Unknown root name produces an error message, does *not* change directory |
| `test_agentic_repo_listing_aggregates_all_roots` | `/repo` shows repos from both roots |
| `test_agentic_repo_callback_encodes_root_name` | Keyboard buttons use a name, not a numeric index |

---

### Backward compatibility checklist

| Scenario | Result |
|---|---|
| Only `APPROVED_DIRECTORY` set | Behaves exactly as before — single root, all output unchanged |
| Both vars set | `APPROVED_DIRECTORIES` wins; `APPROVED_DIRECTORY` acts as final fallback |
| Old `approved_directory=` kwarg to `SecurityValidator` | Accepted, populates `approved_directories[0]` |
| Old `approved_directory=` kwarg to `load_project_registry` | Accepted |
| Old `approved_directory=` kwarg to `check_bash_directory_boundary` | Accepted |
| Telegram `/cd`, `/ls`, `/projects`, `/status`, `/new` with one root | Output identical to before (no root-name prefix) |

---
</details>

### [Allow approved directories list to stand alone](https://github.com/RichardAtCT/claude-code-telegram/pull/80/changes/0999d2a207821386ecf6138869327e8f35848077)

Let settings accept APPROVED_DIRECTORIES without requiring a single
root, and backfill approved_directory from the first entry so
downstream code still has a default working directory. Raise a
clear error when neither setting is provided and document the
either/or requirement.

<details>
  <summary><h4>Click to expand</h4></summary>

### 1. Configuration — `src/config/settings.py`

This is the core logic change.

#### 1a. `approved_directory` is now optional

```python
approved_directory: Optional[Path] = Field(
    None, description="Base directory for projects"
)
```

Why: we want settings to be valid when only `APPROVED_DIRECTORIES` is set.

#### 1b. Validator allows `None`

```python
if v is None:
    return None
```

This keeps validation strict for real paths, but no longer rejects missing
values before we can fall back to the list.

#### 1c. Backfill after parsing

```python
self._approved_directories_cache = self._parse_approved_directories()
if self.approved_directory is None:
    self.approved_directory = self._approved_directories_cache[0]
```

Why: a lot of downstream code assumes a singular default root. This makes that
true again after the multi-root list is parsed and validated.

#### 1d. Fallback now errors if both settings are missing

```python
if self.approved_directory is None:
    raise ValueError(
        "Either APPROVED_DIRECTORY or APPROVED_DIRECTORIES must be set"
    )
return [self.approved_directory]
```

Why: prevents a silent `None` in the approved directory list. The error appears
at settings construction time.

#### 1e. New helper for non-optional usage

```python
def approved_directory_path(self) -> Path:
    if self.approved_directory is None:
        raise ValueError(...)
    return self.approved_directory
```

Why: call sites that need a `Path` can use this property without changing
their own type handling.

#### What to check

- `_parse_approved_directories()` still validates each directory path.
- Backfill happens only after the cache is built.
- The error message is clear and consistent with the new requirement.

---

### 2. Application wiring — `src/main.py`

Two places now use the non-optional helper:

```python
default_working_directory=config.approved_directory_path
```

Why: these paths are required for runtime behavior; if settings are invalid,
we want an immediate, explicit error rather than a `None` propagation.

---

#### New test

```python
test_approved_directories_without_single_directory
```

Ensures that when only the list is provided:

- `approved_directories` parses correctly.
- `approved_directory` is backfilled to the first entry.

`_env_file=None` is passed to avoid the default `.env` loading in tests.

#### Updated fallback test

`test_approved_directories_empty_string_fallback` now also checks that:

- Passing empty `approved_directories_str` **and** `approved_directory=None`
  raises a validation error with the new message.

#### What to check

- The new test asserts both the list and the backfilled single value.
- The error path is exercised and message matches the new requirement.

---

### 4. Documentation — `.env.example`

The comment now states that either `APPROVED_DIRECTORY` or
`APPROVED_DIRECTORIES` is required.

#### What to check

- The comment reflects the new behavior (no fallback requirement).

</details>

### DIFF
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -1464,19 +1464,30 @@ class MessageOrchestrator:
         await query.answer()
 
         data = query.data
-        _, root_name, project_name = data.split(":", 2)
+        parts = data.split(":", 2)
         approved_roots = self.settings.approved_directories
-        # Find the root whose directory name matches what was encoded at render time.
-        # Using names instead of positional indexes means a reordering of
-        # APPROVED_DIRECTORIES between keyboard render and button press cannot
-        # silently route to the wrong root.
-        base = next((r for r in approved_roots if r.name == root_name), None)
-        if base is None:
-            await query.edit_message_text(
-                "Directory root not found. Run /repo again.",
-                parse_mode="HTML",
+        if len(parts) == 3:
+            # New format: cd:<root_name>:<project_name>
+            _, root_name, project_name = parts
+            # Find the root whose directory name matches what was encoded at render time.
+            # Using names instead of positional indexes means a reordering of
+            # APPROVED_DIRECTORIES between keyboard render and button press cannot
+            # silently route to the wrong root.
+            base = next((r for r in approved_roots if r.name == root_name), None)
+            if base is None:
+                await query.edit_message_text(
+                    "Directory root not found. Run /repo again.",
+                    parse_mode="HTML",
+                )
+                return
+        else:
+            # Legacy format: cd:<project_name> (pre-multi-directory keyboards)
+            # Search all approved roots for the named directory; fall back to first root.
+            _, project_name = parts
+            base = next(
+                (r for r in approved_roots if (r / project_name).is_dir()),
+                approved_roots[0],
             )
-            return
         new_path = (base / project_name).resolve()
 
         if not new_path.is_dir():
--- a/src/claude/sdk_integration.py
+++ b/src/claude/sdk_integration.py
@@ -67,7 +67,7 @@ class StreamUpdate:
 def _make_can_use_tool_callback(
     security_validator: SecurityValidator,
     working_directory: Path,
-    approved_directory: Path,
+    approved_directories: List[Path],
 ) -> Any:
     """Create a can_use_tool callback for SDK-level tool permission validation.
 
@@ -107,7 +107,9 @@ def _make_can_use_tool_callback(
             command = tool_input.get("command", "")
             if command:
                 valid, error = check_bash_directory_boundary(
-                    command, working_directory, approved_directory
+                    command,
+                    working_directory,
+                    approved_directories=approved_directories,
                 )
                 if not valid:
                     logger.warning(
@@ -224,7 +226,7 @@ class ClaudeSDKManager:
                 options.can_use_tool = _make_can_use_tool_callback(
                     security_validator=self.security_validator,
                     working_directory=working_directory,
-                    approved_directory=self.config.approved_directory,
+                    approved_directories=self.config.approved_directories,
                 )
 
             # Resume previous session if we have a session_id
--- a/tests/unit/test_bot/test_repo_multi_root.py
+++ b/tests/unit/test_bot/test_repo_multi_root.py
@@ -180,3 +180,83 @@ async def testagentic_repo_callback_encodes_root_name(
         assert not parts[
             1
         ].isdigit(), f"Root identifier should be a name, not an index: {cd!r}"
+
+
+# ---------------------------------------------------------------------------
+# Legacy callback format compatibility shim
+# ---------------------------------------------------------------------------
+
+
+async def _invoke_agentic_callback(settings, query_data: str, context_user_data=None):
+    """Helper: invoke _agentic_callback with a synthetic query."""
+    from src.bot.orchestrator import MessageOrchestrator
+
+    orchestrator = MagicMock(spec=MessageOrchestrator)
+    orchestrator.settings = settings
+    orchestrator._agentic_callback = MessageOrchestrator._agentic_callback.__get__(
+        orchestrator
+    )
+
+    query = AsyncMock()
+    query.data = query_data
+    query.edit_message_text = AsyncMock()
+    query.answer = AsyncMock()
+
+    update = MagicMock()
+    update.callback_query = query
+    update.effective_user.id = 1
+
+    context = MagicMock()
+    context.user_data = context_user_data if context_user_data is not None else {}
+    context.bot_data = {"claude_integration": None}
+
+    await orchestrator._agentic_callback(update, context)
+    return query, context
+
+
+async def test_legacy_callback_format_routes_to_matching_root(
+    dual_root_settings, tmp_path: Path
+):
+    """Old cd:<name> format finds the directory in the correct approved root."""
+    settings, root1, root2 = dual_root_settings
+    # project_a lives in root1, blog lives in root2
+
+    query, context = await _invoke_agentic_callback(settings, query_data="cd:blog")
+
+    assert context.user_data.get("current_directory") == (root2 / "blog").resolve()
+    query.edit_message_text.assert_called_once()
+    confirmation = query.edit_message_text.call_args[0][0]
+    assert "blog" in confirmation
+    assert "not found" not in confirmation.lower()
+
+
+async def test_legacy_callback_format_falls_back_to_first_root_when_not_found(
+    dual_root_settings, tmp_path: Path
+):
+    """Old cd:<name> with an unknown dir falls back to first root (is_dir check catches it)."""
+    settings, root1, root2 = dual_root_settings
+
+    query, context = await _invoke_agentic_callback(
+        settings, query_data="cd:nonexistent"
+    )
+
+    # Should show directory-not-found, not crash
+    query.edit_message_text.assert_called_once()
+    call_text = query.edit_message_text.call_args[0][0]
+    assert "not found" in call_text.lower()
+    assert "current_directory" not in context.user_data
+
+
+async def test_new_callback_format_still_works(dual_root_settings, tmp_path: Path):
+    """New cd:<root_name>:<name> format routes correctly — not broken by the shim."""
+    settings, root1, root2 = dual_root_settings
+
+    query, context = await _invoke_agentic_callback(
+        settings, query_data=f"cd:{root1.name}:project_a"
+    )
+
+    assert context.user_data.get("current_directory") == (root1 / "project_a").resolve()
+    query.edit_message_text.assert_called_once()
+    confirmation = query.edit_message_text.call_args[0][0]
+    assert "project_a" in confirmation
+    assert "not found" not in confirmation.lower()
--- a/tests/unit/test_claude/test_sdk_integration.py
+++ b/tests/unit/test_claude/test_sdk_integration.py
@@ -726,7 +726,7 @@ class TestCanUseToolCallback:
         return _make_can_use_tool_callback(
             security_validator=security_validator,
             working_directory=working_dir,
-            approved_directory=approved_dir,
+            approved_directories=[approved_dir],
         )
 
     @pytest.fixture


RichardAtCT:
Hi @lucianghinda, this is great work — the multi-directory support, test coverage (+440 lines), and backward compatibility are all excellent.

However, there's a significant conflict with the current codebase. Since this PR was created, we've merged PR #62 which **replaced the `ToolMonitor` class** with the SDK's native `can_use_tool` callback (in `src/claude/sdk_integration.py`). The old `ToolMonitor` class no longer exists.

This means your changes to `monitor.py` and the `TestToolMonitor*` tests need to be adapted to the new pattern. Specifically:

1. **`check_bash_directory_boundary()`** still exists in `monitor.py` as a standalone function — your multi-directory extension of this function should rebase cleanly
2. **File path validation** now happens in `_make_can_use_tool_callback()` in `sdk_integration.py` instead of `ToolMonitor.validate_tool_call()`
3. **Tests** for tool validation are now in `tests/unit/test_claude/test_sdk_integration.py` under `TestCanUseToolCallback`

The Settings changes (`APPROVED_DIRECTORIES`), validators, and bot handler updates should rebase without issues.

Would you like to rebase against the current main? Happy to answer any questions about the new `can_use_tool` callback pattern. The key entry point is `_make_can_use_tool_callback()` in `src/claude/sdk_integration.py`.

lucianghinda:
@RichardAtCT thank you for taking the time to review the PR.

Yes, I will rebase from main and see how can I move my changes to fit what's now in main.  It will be a good occasion for me (now with the help of Claude Code) to try to learn more Python along the way (last time I worked with Python  was in 2006 with version 2) while I try to understand the `can_use_tool` callback pattern.

lucianghinda:
@RichardAtCT I have rebased and tried to adapt the changes to be consistent with what is now in main. 

There was a possible breaking change that was flagged by CC: 

> So orchestrator.py:1208 uses data.split(":", 2) and unpacks into three values. If an old cd:myapp button is pressed after the upgrade, split(":", 2) yields ["cd", "myapp"] — only two elements — and  the tuple unpacking raises a ValueError.

I choose to support backward compatibility and put an extra code to detect the old format in case a user is already on the Telegram and tried to answer a question that has the old format while upgrading to the new format. Not 100% sure this case is important but I do want to make sure that any upgrade will not break any flow. 

Here is the thought process that Claude Code used to guide me through it:

<details>
   <summary><h3><strong>Last Commit Explained</strong></h3> </summary> 

This commit is a post-rebase fixup that ties the multi-directory feature into the
security layer. It touches four files across two separate concerns.

---

### Background: what is `can_use_tool`?

Before reviewing the diff, it helps to understand where `can_use_tool` comes from and why it exists.

#### The old approach: `ToolMonitor`

The original codebase had a class called `ToolMonitor`. After each SDK response came back, `ToolMonitor` would inspect what tools Claude had already used and decide whether to raise an error. This is **reactive** — the tool has already run by the time you find out it was problematic.

#### The new approach: `can_use_tool` (merged in PR #62)

`can_use_tool` is a hook provided by `claude-agent-sdk`. You register an async callback on `ClaudeAgentOptions` before the session starts:

```python
options.can_use_tool = my_async_callback
```

The SDK calls this callback **before executing each tool**, passing the tool name and its full input. Returning `PermissionResultAllow()` lets the tool run; returning `PermissionResultDeny(message=...)` blocks it and feeds the denial reason back to Claude so it can adjust.

This is **preventive** rather than reactive: the filesystem is never touched if the path is outside the boundary.

Using `can_use_tool` also required switching from the `async with ClaudeSDKClient(...)` pattern to the explicit `connect()` / `disconnect()` pattern, because the SDK only accepts `can_use_tool` when using that lower-level API.

#### What the callback actually checks

`_make_can_use_tool_callback` is a factory that closes over the session's security context and returns the callback:

```python
def _make_can_use_tool_callback(
    security_validator: SecurityValidator,
    working_directory: Path,
    approved_directories: List[Path],   # ← changed in this commit
) -> AsyncCallable:
```

For each tool call Claude attempts, the returned `can_use_tool` callback does two things:

**1. File path validation** (for `Write`, `Edit`, `Read`, etc.)

```python
if tool_name in _FILE_TOOLS:
    file_path = tool_input.get("file_path") or tool_input.get("path")
    if _is_claude_internal_path(file_path):
        return PermissionResultAllow()   # ~/.claude/plans/, todos/ — always allowed
    valid, _, error = security_validator.validate_path(file_path, working_directory)
    if not valid:
        return PermissionResultDeny(message=error)
```

`SecurityValidator.validate_path` resolves the path and checks it falls within one of the approved directories. It also blocks access to secrets (`.env`, `.ssh/`, `.pem` files) and path traversal sequences.

**2. Bash boundary enforcement** (for `Bash`, `bash`, `shell`)

```python
if tool_name in _BASH_TOOLS:
    command = tool_input.get("command", "")
    valid, error = check_bash_directory_boundary(
        command, working_directory, approved_directories=approved_directories
    )
    if not valid:
        return PermissionResultDeny(message=error)
```

`check_bash_directory_boundary` parses the shell command with `shlex`, splits on `&&`, `||`, `;`, `|`, and `&` to handle chained commands, then checks every absolute or relative path argument against the approved directory list. Commands like `cat` and `ls` are explicitly whitelisted as read-only and always pass. `cd` and filesystem-modifying commands (`mkdir`, `rm`, `mv`, `cp`, etc.) are checked.

---

### What changed in this commit

#### 1. `src/claude/sdk_integration.py` — two changes

**Factory signature: singular → list**

```python
# Before
def _make_can_use_tool_callback(
    ...
    approved_directory: Path,       # one Path
) -> Any:

# After
def _make_can_use_tool_callback(
    ...
    approved_directories: List[Path],   # list of Paths
) -> Any:
```

The factory previously closed over a single `Path`. Now it closes over the full list, which it passes through to `check_bash_directory_boundary`.

**Call site: reads from `config.approved_directories`**

```python
# Before
options.can_use_tool = _make_can_use_tool_callback(
    security_validator=self.security_validator,
    working_directory=working_directory,
    approved_directory=self.config.approved_directory,   # Optional[Path]
)

# After
options.can_use_tool = _make_can_use_tool_callback(
    security_validator=self.security_validator,
    working_directory=working_directory,
    approved_directories=self.config.approved_directories,  # List[Path]
)
```

`config.approved_directories` is the new list property added by the multi-directory commits. It always returns at least one element — the list is validated at startup — so there is no `None` risk here. This also fixes a latent mypy error: `approved_directory` is `Optional[Path]` and the old code passed it as `Path` without a guard.

#### 2. `src/bot/orchestrator.py` — compatibility shim for `/repo` callbacks

The multi-directory commits changed the inline keyboard callback data format from `cd:<name>` to `cd:<root_name>:<name>`. Without a shim, any user who had a `/repo` keyboard open before the deployment would trigger a `ValueError` when pressing a button, because the old two-part string can't be unpacked into three variables.

The shim detects the format by counting the parts:

```python
parts = data.split(":", 2)
if len(parts) == 3:
    # New format: cd:<root_name>:<project_name>
    _, root_name, project_name = parts
    base = next((r for r in approved_roots if r.name == root_name), None)
    if base is None:
        await query.edit_message_text("Directory root not found. Run /repo again.", ...)
        return
else:
    # Legacy format: cd:<project_name>
    _, project_name = parts
    base = next(
        (r for r in approved_roots if (r / project_name).is_dir()),
        approved_roots[0],
    )
```

For the legacy path: rather than guessing which root the directory belongs to by name alone, the shim searches all approved roots for a directory with that name. If it finds one, it uses that root. If the name doesn't exist in any root, it falls back to the first root — which then fails the `is_dir()` check a few lines later and produces a clean user-visible error rather than a crash.

#### 3. `tests/unit/test_bot/test_repo_multi_root.py` — three new tests

A shared helper `_invoke_agentic_callback` wires up the orchestrator and a synthetic Telegram callback query without touching the real bot runtime.

| Test | What it verifies |
|---|---|
| `test_legacy_callback_format_routes_to_matching_root` | `cd:blog` (old format) finds `blog` in the correct root and navigates there |
| `test_legacy_callback_format_falls_back_to_first_root_when_not_found` | `cd:nonexistent` (old format) produces a "not found" message, does not crash or set `current_directory` |
| `test_new_callback_format_still_works` | `cd:work:project_a` (new format) still routes correctly after the shim was added |

#### 4. `tests/unit/test_claude/test_sdk_integration.py` — one-line change

The `callback` fixture in `TestCanUseToolCallback` was passing `approved_directory=` to
the factory. Updated to `approved_directories=[approved_dir]` to match the new signature:

```python
# Before
return _make_can_use_tool_callback(
    security_validator=security_validator,
    working_directory=working_dir,
    approved_directory=approved_dir,
)

# After
return _make_can_use_tool_callback(
    security_validator=security_validator,
    working_directory=working_dir,
    approved_directories=[approved_dir],
)
```

All existing `TestCanUseToolCallback` tests continue to exercise the same scenarios (allow safe read, deny invalid path, allow bash inside boundary, deny bash outside boundary, etc.) — the wrap in a list is transparent to the test logic.

---

### How all the pieces fit together at runtime

When a user sends a message and Claude starts executing:

```
User message
    │
    ▼
ClaudeSDKManager.execute_command()
    │  builds ClaudeAgentOptions with:
    │    options.can_use_tool = _make_can_use_tool_callback(
    │        security_validator,
    │        working_directory,
    │        approved_directories=["/work/projects", "/work/personal"]
    │    )
    │
    ▼
claude-agent-sdk runs Claude
    │
    ├── Claude wants to run: Write(file_path="/work/projects/app/main.py")
    │       SDK calls can_use_tool("Write", {"file_path": "..."}, ctx)
    │       → security_validator.validate_path() → within /work/projects ✓
    │       → PermissionResultAllow()  →  file written
    │
    ├── Claude wants to run: Write(file_path="/etc/passwd")
    │       SDK calls can_use_tool("Write", {"file_path": "/etc/passwd"}, ctx)
    │       → security_validator.validate_path() → outside all approved dirs ✗
    │       → PermissionResultDeny("Access denied: path outside approved directories")
    │       → Claude receives the denial and knows not to retry
    │
    └── Claude wants to run: Bash(command="rm /tmp/evil && mkdir newdir")
            SDK calls can_use_tool("Bash", {"command": "..."}, ctx)
            → check_bash_directory_boundary() parses and splits on &&
            → "rm /tmp/evil": /tmp is outside approved dirs ✗
            → PermissionResultDeny("Directory boundary violation: ...")
```

</details>

RichardAtCT:
Hi @lucianghinda — heads up that several PRs have been merged to main since this was opened, including Phase 5 cleanup (#96) which heavily refactored `sdk_integration.py` and `monitor.py`, plus #99 (MCP server) and #110/#107/#111 which all touched `orchestrator.py`. This PR will need a substantial rebase before it can be reviewed.

When you're ready to rebase, the key changes to be aware of:
- `find_claude_cli()` was removed — `cli_path` is now passed directly from config
- `ToolMonitor` class was deleted — replaced by `can_use_tool` callback in sdk_integration.py
- `monitor.py` now only contains `check_bash_directory_boundary` and `_is_claude_internal_path`
- `orchestrator.py` gained MCP image handling, progress message error handling, and forum topic routing

Happy to help review once rebased!

lucianghinda:
I resumed working on this today - will rebase and address thre comments.

LE: ✅ done I hope I kept the same code quality and address all the comments

RichardAtCT:
Closing — this PR has been stale since Feb 21 with no response to the rebase request. The codebase has changed significantly since then (ToolMonitor removed, can_use_tool callback pattern). If you'd like to pick this back up, feel free to reopen with a fresh rebase against main. Thanks for the contribution\!

PR #84, opened 2026-02-21
fix: throttle project-thread sync telegram api bursts

### LINKED ISSUE #73: Telegram Bot banned after a few hours
Hi, not urgent, just checking with you.

I initiated the bot, and after a few hours the bot account was deleted by Telegram. Did that ever happen to you?

Could it have been flagged for spam when it initiated my 12 threads in seconds during the config?

drugoi:
## Summary
- keep issue #73 private-chat mitigation by pacing all Telegram API calls in `ProjectThreadManager.sync_topics`
- add global outbound PTB limiter via `AIORateLimiter(max_retries=1)` during bot initialization
- move flood-retry ownership to PTB limiter and remove manager-local `RetryAfter` retry logic to avoid duplicate retries/sleeps
- use `python-telegram-bot[rate-limiter]` dependency wiring (lockfile updated)
- keep safe default `PROJECT_THREADS_SYNC_ACTION_INTERVAL_SECONDS=1.1` when env is absent
- update unit tests for sync pacing/no-local-retry and bot-core limiter configuration

## Why Keep Sync Interval Setting
- `AIORateLimiter` group throttling is keyed for negative `chat_id` (groups/channels); private chats (positive IDs) fall back to global limiting.
- issue #73 is caused by bursty private-chat topic bootstrap traffic, so sync-specific pacing is still needed as a targeted safeguard.
- removing the sync interval would require stricter global limits and could slow unrelated bot traffic.

## Testing
- `poetry run pytest -q tests/unit/test_projects/test_thread_manager.py tests/unit/test_bot/test_thread_mode_handlers.py tests/unit/test_bot/test_core_rate_limiter.py tests/unit/test_config.py`

Closes #73

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -13,6 +13,18 @@ files = [
 ]
 
 [[package]]
+name = "aiolimiter"
+version = "1.2.1"
+description = "asyncio rate limiter, a leaky bucket implementation"
+optional = false
+python-versions = "<4.0,>=3.8"
+groups = ["main"]
+files = [
+    {file = "aiolimiter-1.2.1-py3-none-any.whl", hash = "sha256:d3f249e9059a20badcb56b61601a83556133655c11d1eb3dd3e04ff069e5f3c7"},
+    {file = "aiolimiter-1.2.1.tar.gz", hash = "sha256:e02a37ea1a855d9e832252a105420ad4d15011505512a1a1d814647451b5cca9"},
+]
+
+[[package]]
 name = "aiosqlite"
 version = "0.21.0"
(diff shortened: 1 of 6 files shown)


RichardAtCT:
Did you consider using PTBs built in rate limiter? https://docs.python-telegram-bot.org/en/stable/telegram.ext.aioratelimiter.html

drugoi:
@RichardAtCT if I understood right, AIORateLimiter works only in group/supergroup/channels(?), not in personal chats. So part of the solution is still required custom rate-limiter.

PR #103, opened 2026-02-24
Enable concurrent processing of messages across topics

### LINKED ISSUE #102: Enable concurrent processing of messages across topics
## Summary

When using project threads mode (group), messages sent to different topics are processed sequentially. This means a long-running Claude task in one project topic blocks all other topics until it completes.

## Proposed Change

Enable concurrent message processing so that messages in different project topics are handled in parallel. Each topic/handler gets its own isolated snapshot of thread state (working directory, session ID, thread context) to avoid race conditions.

### Key Design Decisions

- **Per-update state isolation**: `context.user_data` is shared per-user, so concurrent handlers for different topics would overwrite each other's values. Thread state is stored in a per-update dictionary on the orchestrator instance, keyed by `update.update_id`.
- **Note**: `telegram.Update` uses `__slots__`, so arbitrary attributes cannot be set on it directly — an instance-level dictionary is used instead.
- **Claude Code internal paths**: Concurrent processing surfaces more tool-result reads (`~/.claude/projects/`, `/tmp/claude-<uid>/`). These paths need to be whitelisted in `_CLAUDE_INTERNAL_SUBDIRS` alongside `plans` and `todos`.

I have an implementation ready at https://github.com/tschuehly/claude-code-telegram/tree/fix/concurrent-topic-processing and can open a PR.

tschuehly:
## Summary

- Store per-update thread state in a dict on the orchestrator instance instead of setting attributes on `telegram.Update` (which uses `__slots__` and raises `AttributeError`)
- Add `projects` and `tasks` to `_CLAUDE_INTERNAL_SUBDIRS` whitelist so Claude Code can read its own compressed tool results and task data
- Allow `/tmp/claude-<uid>/` paths for Task agent output files
- Add "Viewing Logs" section to README

Closes #102

## Test plan

- [ ] Send messages to multiple project topics concurrently and verify they are processed in parallel
- [ ] Verify plan mode works (writes to `~/.claude/plans/`)
- [ ] Verify long conversations work (Claude re-reads compressed tool results from `~/.claude/projects/`)
- [ ] Verify Task agent subprocesses work (reads from `/tmp/claude-<uid>/`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,21 @@ To use topics with your bot, enable them in BotFather:
 
 Message [@userinfobot](https://t.me/userinfobot) on Telegram -- it will reply with your user ID number.
 
+## Viewing Logs
+
+If running as a systemd user service (see [SYSTEMD_SETUP.md](SYSTEMD_SETUP.md)):
+
+```bash
+# Live logs (follow mode)
+journalctl --user -u claude-telegram-bot -f
+
+# Recent logs (last 50 lines)
+journalctl --user -u claude-telegram-bot -n 50
+
+# Check service status
+systemctl --user status claude-telegram-bot
+```
+
 ## Troubleshooting
 
 **Bot doesn't respond:**
--- a/src/bot/core.py
+++ b/src/bot/core.py
@@ -51,6 +51,7 @@ class ClaudeCodeBot:
         builder = Application.builder()
         builder.token(self.settings.telegram_token_str)
         builder.rate_limiter(AIORateLimiter(max_retries=1))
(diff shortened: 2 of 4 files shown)


RichardAtCT:
Closing — CI has been failing and merge conflicts plus the changes requested in review since February remain unaddressed. The concurrent topic processing design is sound (per-update state isolation is the right approach), but this needs a fresh implementation with proper test coverage and memory leak prevention. See #102 for the tracking issue — contributions welcome with a fresh PR.

PR #106, opened 2026-02-25
Add voice message transcription (Mistral Voxtral + OpenAI Whisper)

guillaumegay13:
## Summary
  - Add voice message transcription support for the Telegram bot
  - Two providers: **Mistral Voxtral** (default) and **OpenAI Whisper**
  - Downloads OGG audio from Telegram, sends to configured API, passes text to Claude
  - Integrated in both agentic and classic mode handlers

  ## Configuration

  | Variable | Default | Description |
  |---|---|---|
  | `VOICE_PROVIDER` | `mistral` | `mistral` or `openai` |
  | `MISTRAL_API_KEY` | — | Required when `VOICE_PROVIDER=mistral` |
  | `OPENAI_API_KEY` | — | Required when `VOICE_PROVIDER=openai` |
  | `VOICE_TRANSCRIPTION_MODEL` | auto | `voxtral-mini-latest` (Mistral) or `whisper-1` (OpenAI) |

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -102,7 +102,7 @@ Multi-project topics: `ENABLE_PROJECT_THREADS` (default false), `PROJECT_THREADS
 
 Output verbosity: `VERBOSE_LEVEL` (default 1, range 0-2). Controls how much of Claude's background activity is shown to the user in real-time. 0 = quiet (only final response, typing indicator still active), 1 = normal (tool names + reasoning snippets shown during execution), 2 = detailed (tool names with input summaries + longer reasoning text). Users can override per-session via `/verbose 0|1|2`. A persistent typing indicator is refreshed every ~2 seconds at all levels.
 
-Voice transcription: `ENABLE_VOICE_MESSAGES` (default true), `VOICE_PROVIDER` (`mistral`|`openai`, default `mistral`), `MISTRAL_API_KEY`, `OPENAI_API_KEY`, `VOICE_TRANSCRIPTION_MODEL`. Provider implementations in `src/bot/voice/`.
+Voice transcription: `ENABLE_VOICE_MESSAGES` (default true), `VOICE_PROVIDER` (`mistral`|`openai`, default `mistral`), `MISTRAL_API_KEY`, `OPENAI_API_KEY`, `VOICE_TRANSCRIPTION_MODEL`. Provider implementation is in `src/bot/features/voice_handler.py`.
 
 Feature flags in `src/config/features.py` control: MCP, git integration, file uploads, quick actions, session export, image uploads, voice messages, conversation mode, agentic mode, API server, scheduler.
 
--- a/src/bot/features/voice_handler.py
+++ b/src/bot/features/voice_handler.py
@@ -2,7 +2,7 @@
 
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Optional
+from typing import Any, Optional
 
 import structlog
 from telegram import Voice
@@ -26,6 +26,8 @@ class VoiceHandler:
 
     def __init__(self, config: Settings):
         self.config = config
+        self._mistral_client: Optional[Any] = None
+        self._openai_client: Optional[Any] = None
 
     def _ensure_allowed_file_size(self, file_size: Optional[int]) -> None:
         """Reject files that exceed the configured max size."""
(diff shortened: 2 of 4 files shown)


PeterUstinox:
very nice! i was thinking the same 
@RichardAtCT this feature is not mentioned in the  [CONTRIBUTING.md](https://github.com/RichardAtCT/claude-code-telegram/blob/main/CONTRIBUTING.md) but i would love to see it! 

Any chance of putting this in 1.4?

RichardAtCT:
Love it! Can you please fix the failing CI and then I'll test and look to merge

RichardAtCT:
Also can you do a doc update as well?

guillaumegay13:
hi @RichardAtCT, done in the latest commit. Thank you!

guillaumegay13:
Thanks - I addressed those points in the latest commit: 

Blocking fixes:
- Make mistralai/openai optional extras (already in pyproject.toml);
  regenerate poetry.lock so markers are in sync for release installs (P1)
- Add VOICE_MAX_FILE_SIZE_MB setting with configurable limit (default
  20 MB) checked before downloading voice data
- Provider-aware fallback error messages in both agentic and classic
  handlers (reference correct API key env var per provider)
- Validate voice_provider with Literal["mistral", "openai"] + field
  validator; invalid values like "google" now raise ValueError at startup
- Use lazy imports for mistralai/openai with pip-friendly install hint

Non-blocking fixes:
- Extract _handle_agentic_media_message shared helper to deduplicate
  agentic_voice and agentic_photo
- Add tests for transcription API failures (network error, empty
  response) and missing optional dependency
- Add orchestrator-level tests for voice path and provider-aware
  error messages
- Add classic handler tests for voice fallback messages
- Document VOICE_MAX_FILE_SIZE_MB in configuration.md (P3)
- Document voice extras install for pip/uv users in setup.md (P2)

guillaumegay13:
New commit to address issues and questions, @FridayOpenClawBot please review!

RichardAtCT:
> **PR Review**
> Reviewed head: `cc1e94499ef2e3a0ef859c48d9f38fe9d6bf99d4`
> 
> **Summary**
> - Adds voice message transcription support with dual providers: Mistral Voxtral (default) and OpenAI Whisper
> - Downloads Telegram voice (.ogg), transcribes via configured provider, then passes transcribed text to Claude
> - Integrated in both agentic and classic modes; packaged as an optional `[voice]` extras group
> 
> **What looks good**
> - Triple file-size check (initial metadata → resolved metadata → post-download byte count) is thorough and prevents unbounded downloads
> - Unknown-size payloads are rejected proactively rather than silently downloaded — good defensive practice
> - Optional dependency import with helpful error messages (missing `[voice]` extras) is user-friendly
> - Good test coverage: 354 lines in `test_voice_handler.py` + 55 in `test_message_voice.py`
> 
> **Issues / questions**
> 
> 1) **[Important]** `src/bot/features/voice_handler.py` (`process_voice_message`) — The transcribed text is passed directly as a prompt to Claude without any sanitization or length check. A malicious actor could craft a voice message that, when transcribed, contains prompt injection instructions (e.g., "Ignore previous instructions and…"). Since Telegram users are presumably authenticated/authorized, this may be acceptable, but it's worth documenting the trust model. At minimum, consider capping prompt length or stripping obvious injection patterns.
> 
> 2) **[Important]** `src/bot/features/voice_handler.py` — The `VoiceHandler` instance holds cached `_mistral_client` / `_openai_client` objects that persist across requests. If API keys are rotated at runtime (unlikely but possible), the cached clients will continue using stale keys until the bot restarts. A `reset()` method or config-hash-based invalidation would be prudent for long-running deployments.
> 
> 3) **[Important]** `src/config/settings.py` (implied) — The docs say `MISTRAL_API_KEY` is required when `VOICE_PROVIDER=mistral` and vice versa for OpenAI — but is this validated at startup (fail-fast) or only at first voice message? If validation is deferred, an operator misconfiguration won't surface until a user sends a voice message. Fail-fast validation in `settings.py` would be safer.
> 
> 4) **[Nit]** `src/bot/features/voice_handler.py` (`process_voice_message`) — `caption if caption else "Voice message transcription:"` will use the default label for empty-string captions too. `caption or "Voice message transcription:"` is shorter and more idiomatic.
> 
> **Suggested tests (if needed)**
> - Test that a voice handler with a stale cached client still works after key rotation (or document it as not supported)
> - Test for very long transcriptions (e.g., >10k chars) to ensure the prompt passed to Claude is within reasonable bounds
> 
> **Verdict**
> ⚠️ Merge after fixes — item 1 (prompt injection via transcription) warrants at least a documented decision. Items 2 and 3 are operational quality issues worth addressing. Otherwise the implementation is solid.
> 
> — *Friday, AI assistant to @RichardAtCT*
> 

Not sure I agree with point 1! By definition any message could've a prompt injection - not sure I think it is necessary to consider.

guillaumegay13:
Agree with @RichardAtCT on (1). I’d push back on (2) and (3): for (2), runtime API key rotation is a pretty rare ops scenario, and if a cached client has a stale key it fails safely until restart/refresh; for (3), this is intentional deferred validation. VOICE_PROVIDER is validated at startup, while missing provider keys only disable voice (with a provider-aware message) instead of blocking the whole bot from starting.

guillaumegay13:
@RichardAtCT should we merge it?

RichardAtCT (approved):
Previous blocking feedback addressed — deps are now optional. LGTM, ready to merge.

PR #112, opened 2026-02-26
feat: add /restart command

F1orian:
## Summary
- Adds a `/restart` Telegram command that gracefully restarts the bot process
- Sends a confirmation message, then triggers `SIGTERM` so systemd (or any process manager with `Restart=always`) brings the bot back up
- Registered in both agentic and classic handler modes, with audit logging
- **Fixes a bug where the command menu was never pushed to Telegram** — `set_my_commands()` was called before the Application's HTTP client was initialized (see below)

## Bug fix: command menu not updating

`_set_bot_commands()` in `ClaudeCodeBot.initialize()` was called right after `builder.build()`, but `build()` only constructs the `Application` object — it does **not** initialize the bot's HTTP client. The actual `app.initialize()` (which creates the HTTPX async client and calls `getMe()`) happened much later in `start()`.

This meant `set_my_commands()` was hitting the Telegram API with no HTTP client ready, so the command menu was never actually pushed. Handlers still worked because they're just Python objects added to internal lists.

**Fix:** Call `await self.app.initialize()` in the bot's `initialize()` method, before `_set_bot_commands()`. The later `app.initialize()` call in `start()` is idempotent (ptb checks `_initialized`), so there's no double-init issue. This also fixes command menu updates for any commands added in the future.

## Test plan
- [x] Send `/restart` in Telegram chat
- [x] Verify bot replies "🔄 Restarting bot…"
- [x] Verify bot goes offline briefly and comes back within ~10 seconds
- [x] Check `journalctl` for clean shutdown + restart cycle
- [x] Verify command appears in Telegram's `/` command menu

### DIFF
--- a/src/bot/handlers/command.py
+++ b/src/bot/handlers/command.py
@@ -1228,12 +1228,16 @@ async def restart_command(update: Update, context: ContextTypes.DEFAULT_TYPE) ->
 
     Sends a confirmation message then triggers SIGTERM so systemd
     (or any process manager with restart-on-exit) brings the bot back up.
+
+    Auth: protected by the auth middleware (group -2) which raises
+    ``ApplicationHandlerStop`` for unauthenticated users before any
+    handler in group 10 runs.  No per-handler check is needed.
     """
     audit_logger: AuditLogger = context.bot_data.get("audit_logger")
     user_id = update.effective_user.id
 
     await update.message.reply_text(
-        "🔄 <b>Restarting bot…</b>\n\nBack in a few seconds.",
+        "🔄 <b>Restarting bot…</b>\n\nBack shortly.",
         parse_mode="HTML",
     )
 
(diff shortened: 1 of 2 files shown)


F1orian:
Thanks for the review! Pushed a fix commit addressing both points:

**1. Auth concern (confirmed safe):**
The auth middleware runs at handler group `-2` and raises `ApplicationHandlerStop` for unauthenticated users — this blocks **all** handlers at group `10` (where `/restart` lives) before they execute. There is no registration path that bypasses this middleware chain. Added a docstring note to `restart_command` documenting this explicitly.

**2. Wording nit:**
Changed "Back in a few seconds" → "Back shortly" since `RestartSec` / process manager timing isn't guaranteed.

**3. Suggested test added:**
Added `test_restart_command_sends_sigterm` — mocks `os.kill` and verifies it's called with `(os.getpid(), signal.SIGTERM)`, plus checks the confirmation message is sent.

> Unit test that an unauthorized user cannot trigger /restart

This is already covered by the existing middleware test suite (`test_auth_rejection_raises_handler_stop`, `test_real_auth_middleware_rejection` in `test_middleware.py`) which proves that *any* unauthenticated update is blocked before reaching any handler. No per-command auth test is needed.

PR #114, opened 2026-02-26
fix: pass allowed_tools=None to SDK when DISABLE_TOOL_VALIDATION=true

### LINKED ISSUE #91: DISABLE_TOOL_VALIDATION doesn't affect SDK-level allowed_tools, blocking MCP tools
## Problem

When using MCP servers (e.g., Home Assistant MCP), all `mcp__*` tool calls are blocked because:

1. `DISABLE_TOOL_VALIDATION=true` only skips the bot-side `ToolMonitor` checks in `monitor.py`
2. The hardcoded default `claude_allowed_tools` list (17 tools) is still passed to `ClaudeAgentOptions(allowed_tools=...)` in `sdk_integration.py:177`
3. The Claude Code CLI enforces this list independently, blocking any tool not in the default list

This means MCP tools like `mcp__home-assistant__ha_get_overview` are always blocked, regardless of `DISABLE_TOOL_VALIDATION`.

## Why `CLAUDE_ALLOWED_TOOLS` doesn't help

The env var uses exact string matching (`tool_name not in list`), so prefix patterns like `mcp__home-assistant` don't match `mcp__home-assistant__ha_get_overview`. Users would need to list every individual MCP tool name, which is impractical since MCP servers can expose dozens of tools and the list changes when the server is updated.

## Expected behavior

When `DISABLE_TOOL_VALIDATION=true`, the SDK should receive `allowed_tools=None` (allow all) instead of the hardcoded default list. This would let the Claude Code CLI allow all tools, including MCP tools.

## Suggested fix

In `sdk_integration.py`, check `disable_tool_validation` before passing the list:

```python
options = ClaudeAgentOptions(
    allowed_tools=None if self.config.disable_tool_validation else self.config.claude_allowed_tools,
    disallowed_tools=None if self.config.disable_tool_validation else self.config.claude_disallowed_tools,
    ...
)
```

Alternatively, support wildcard/prefix matching in `CLAUDE_ALLOWED_TOOLS` (e.g., `mcp__home-assistant` matching all tools from that server).

## Workaround

We're currently patching `sdk_integration.py` in our container entrypoint to replace `allowed_tools=self.config.claude_allowed_tools` with `allowed_tools=None` when `DISABLE_TOOL_VALIDATION=true`.

sogadaiki:
## Summary

- When `DISABLE_TOOL_VALIDATION=true`, pass `allowed_tools=None` and `disallowed_tools=None` to `ClaudeAgentOptions` so the SDK does not restrict tool usage
- MCP tools (e.g. `mcp__home-assistant__*`) are no longer blocked at the SDK level when tool validation is disabled
- The `can_use_tool` callback continues to enforce file path and bash directory boundaries regardless

Closes #91

## Test plan

- [x] New test: `test_disable_tool_validation_sets_allowed_tools_none` — verifies `allowed_tools=None` when flag is true
- [x] New test: `test_tool_validation_enabled_passes_configured_tools` — verifies configured lists are passed when flag is false (default)
- [x] All 405 existing tests pass
- [ ] Manual test: enable MCP server, set `DISABLE_TOOL_VALIDATION=true`, verify MCP tools are accessible

### DIFF
--- a/src/claude/sdk_integration.py
+++ b/src/claude/sdk_integration.py
@@ -172,12 +172,22 @@ class ClaudeSDKManager:
                 logger.debug("Claude CLI stderr", line=line)
 
             # Build Claude Agent options
+            # When DISABLE_TOOL_VALIDATION=true, pass None for allowed/disallowed
+            # tools so the SDK does not restrict tool usage. This lets MCP tools
+            # (e.g. mcp__home-assistant__*) through at the SDK level.
+            if self.config.disable_tool_validation:
+                sdk_allowed_tools = None
+                sdk_disallowed_tools = None
+            else:
+                sdk_allowed_tools = self.config.claude_allowed_tools
+                sdk_disallowed_tools = self.config.claude_disallowed_tools
+
             options = ClaudeAgentOptions(
                 max_turns=self.config.claude_max_turns,
                 max_budget_usd=self.config.claude_max_cost_per_request,
                 cwd=str(working_directory),
-                allowed_tools=self.config.claude_allowed_tools,
-                disallowed_tools=self.config.claude_disallowed_tools,
+                allowed_tools=sdk_allowed_tools,
+                disallowed_tools=sdk_disallowed_tools,
                 cli_path=self.config.claude_cli_path,
                 sandbox={
                     "enabled": self.config.sandbox_enabled,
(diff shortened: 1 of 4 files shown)


RichardAtCT:
Applied the core fix onto main (commit 36464a3) — stripped the unrelated bot_language test fixture changes. Thanks for the contribution! Closes #91.

PR #115, opened 2026-02-26
fix: treat Topic_not_modified as success in topic sync

### LINKED ISSUE #108: Topic sync treats already-open topics as failures (Topic_not_modified)
## Bug

On startup, `ProjectThreadManager.sync_topics()` calls `reopenForumTopic` to verify each topic is usable. When a topic is **already open**, the Telegram API returns `400 Bad Request` with `Topic_not_modified`. This is not recognized as a success case — it falls through to the generic `TelegramError` handler and is counted as a failure.

**Result:** All project topics show as `failed` on every startup even though they are perfectly usable.

```
{"mode": "group", "chat_id": ..., "created": 0, "reused": 0, "renamed": 0, "failed": 4, "deactivated": 0, "event": "Project thread startup sync complete"}
```

## Root Cause

In `src/projects/thread_manager.py`, both `_ensure_topic_usable()` and `_reopen_topic_if_possible()` catch `TelegramError` and check `_is_topic_unusable_error()`. The string `"topic_not_modified"` is not in the marker list, so it falls through to `return "failed"`.

## Suggested Fix

Add a check for the "already open" case before the unusable check:

```python
@staticmethod
def _is_topic_already_open(error: TelegramError) -> bool:
    """Return True when the topic is already open (not modified)."""
    text = str(error).lower()
    return "topic_not_modified" in text or "topic not modified" in text
```

Then in `_ensure_topic_usable()`:

```python
except TelegramError as e:
    if self._is_topic_already_open(e):
        return "ok"
    if self._is_topic_unusable_error(e):
        return "unusable"
    ...
```

And similarly in `_reopen_topic_if_possible()`, return `"reopened"` for the already-open case.

## After Fix

```
{"mode": "group", "chat_id": ..., "created": 0, "reused": 4, "renamed": 0, "failed": 0, "deactivated": 0, "event": "Project thread startup sync complete"}
```

## Environment

- claude-code-telegram v1.3.0
- python-telegram-bot (Telegram API returns `BadRequest("Topic_not_modified")` for HTTP 400)

sogadaiki:
## Summary

- Add `_is_topic_already_open()` to detect `Topic_not_modified` errors from Telegram API
- In `_ensure_topic_usable()`: treat already-open topics as `"ok"` instead of `"failed"`
- In `_reopen_topic_if_possible()`: treat already-open topics as `"reopened"` instead of `"failed"`
- On startup, existing open topics now correctly count as `reused` instead of `failed`

Closes #108

## Test plan

- [x] New test: `test_ensure_topic_usable_already_open_returns_ok` — verifies `Topic_not_modified` error returns "ok"
- [x] New test: `test_sync_active_mapping_already_open_topic_counts_as_reused` — verifies sync result shows `reused=2, failed=0`
- [x] All 405 existing tests pass (13 existing thread manager tests + 2 new)
- [ ] Manual test: restart bot with existing open topics, verify log shows `reused: N, failed: 0`

### DIFF
--- a/src/projects/thread_manager.py
+++ b/src/projects/thread_manager.py
@@ -263,6 +263,8 @@ class ProjectThreadManager:
             )
             return "ok"
         except TelegramError as e:
+            if self._is_topic_already_open(e):
+                return "ok"
             if self._is_topic_unusable_error(e):
                 return "unusable"
             logger.warning(
@@ -286,6 +288,8 @@ class ProjectThreadManager:
             )
             return "reopened"
         except TelegramError as e:
+            if self._is_topic_already_open(e):
+                return "reopened"
             if self._is_topic_unusable_error(e):
                 return "unusable"
             logger.warning(
@@ -407,6 +411,16 @@ class ProjectThreadManager:
                 project_name=project_name,
                 error=str(e),
             )
+
+    @staticmethod
+    def _is_topic_already_open(error: TelegramError) -> bool:
+        """Return True when topic is already open (reopen is a no-op)."""
+        text = str(error).lower()
+        markers = [
+            "topic_not_modified",
+            "topic not modified",
+        ]
+        return any(marker in text for marker in markers)
 
     @staticmethod
     def _is_topic_unusable_error(error: TelegramError) -> bool:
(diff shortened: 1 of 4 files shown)


sogadaiki:
Closing — the maintainer applied an equivalent fix directly in 4b3d101. Issue #108 is already closed. Thanks!

PR #116, opened 2026-02-26
feat: add /stop command to cancel running Claude calls

F1orian:
## Summary
- Adds `/stop` command that immediately cancels a running Claude call
- Enables `concurrent_updates(True)` on the PTB Application so `/stop` can run while a long-running Claude call is in progress (PTB defaults to sequential update processing, which blocked `/stop` behind the active handler)
- Tracks active calls per-thread in `_active_calls` dict keyed by thread state_key, so `/stop` in one project thread doesn't affect another
- Replaces single `_active_pid` with per-call PID tracking (`_active_pids` dict) and process-tree killing via `/proc/<pid>/task/<pid>/children` to safely terminate the CLI and all its subagents without killing the bot's process group
- Registered in both agentic and classic modes with bot menu entry "Stop running Claude call"

## Changes
- `src/bot/core.py` — enable `concurrent_updates(True)`
- `src/bot/orchestrator.py` — per-thread task tracking via `_active_calls`, `_thread_key()` helper
- `src/bot/handlers/message.py` — same per-thread tracking for classic mode
- `src/bot/handlers/command.py` — `/stop` resolves correct thread, uses `abort_call(call_id)`
- `src/claude/facade.py` — thread `call_id` through `run_command`, add `abort_call()`
- `src/claude/sdk_integration.py` — per-call PID dict, process-tree kill (`_get_descendants`, `_kill_pid`), `abort()` for shutdown, `abort_call()` for targeted stop

## Test plan
- [x] Updated orchestrator tests for new command counts (agentic: 5→6, classic: 13→14)
- [x] All orchestrator tests pass (9/9)
- [x] Manual test: `/stop` cancels running call immediately (confirmed via logs: `success: true`)
- [x] Manual test: `/stop` with nothing running → "Nothing running." response
- [x] Manual test: concurrent calls in two project threads, `/stop` in one does not affect the other
- [x] Manual test: bot stays alive after `/stop` (no process group kill)
- [x] Subagent processes terminated via process-tree walking

### DIFF
--- a/src/bot/features/interactive_questions.py
+++ b/src/bot/features/interactive_questions.py
@@ -89,7 +89,11 @@ def build_single_select_keyboard(
         )
     rows = [buttons[i : i + 2] for i in range(0, len(buttons), 2)]
     rows.append(
-        [InlineKeyboardButton(text="Other...", callback_data=f"askq:{question_idx}:other")]
+        [
+            InlineKeyboardButton(
+                text="Other...", callback_data=f"askq:{question_idx}:other"
+            )
+        ]
     )
     return InlineKeyboardMarkup(rows)
 
@@ -110,7 +114,11 @@ def build_multi_select_keyboard(
         )
     rows = [buttons[i : i + 2] for i in range(0, len(buttons), 2)]
     rows.append(
-        [InlineKeyboardButton(text="Other...", callback_data=f"askq:{question_idx}:other")]
+        [
+            InlineKeyboardButton(
+                text="Other...", callback_data=f"askq:{question_idx}:other"
+            )
+        ]
     )
     rows.append(
         [InlineKeyboardButton(text="Done ✓", callback_data=f"askq:{question_idx}:done")]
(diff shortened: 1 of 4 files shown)


F1orian:
Thanks for the thorough review! Pushed fixes for all three points:

**1. `context.user_data` race (confirmed safe + documented):**
All reads from `user_data` (`_thread_context`, `_active_calls`, `task`, `call_id`) happen synchronously before the first `await` in `stop_command`, so no concurrent handler can mutate the dict between read and cancel. Added a comment documenting this invariant.

**2. Linux-only `/proc` process-tree walking:**
Added a `platform.system() != "Linux"` guard in `_get_descendants()` — returns `[]` on non-Linux so `_kill_pid` gracefully falls back to killing only the root process. Docstring updated.

**3. `id(object())` → monotonic counter:**
Replaced with `itertools.count(1)` in both `orchestrator.py` (class-level `_call_counter`) and `message.py` (module-level `_call_counter`). No address-reuse risk.

**Tests added:**
- `test_stop_command_nothing_running` — verifies "Nothing running." reply when `_active_calls` is empty
- `test_stop_command_thread_isolation` — verifies `/stop` in thread A cancels only thread A's task, leaving thread B untouched

RichardAtCT:
Just catching up with my clanker! 
Point 1 on the latest comment needs to be addressed because a user can run the tool in production on MacOS (I do!).

RichardAtCT:
Closing — superseded by #122 which implements the same functionality (cancel running requests) via an inline Stop button rather than a /stop command. The inline approach provides better UX and a smaller diff. Thanks for the contribution!

PR #123, opened 2026-03-03
feat: stream partial responses via Telegram sendMessageDraft API

nimdraugsael:
Recently Telegram API released support of response streaming for Telegram bots. Attached feature preview from Durov's telegram channel

https://github.com/user-attachments/assets/5dc015db-13fd-4222-8e68-56a6fef237fc

DraftStreamer accumulates two sections — a tool header (showing tool calls and reasoning snippets as they arrive) and a response body (streamed token-by-token) — and sends throttled drafts that self-disable on API errors for graceful fallback.

SDK changes:
- Enable --include-partial-messages for token-level StreamEvent deltas (content_block_delta / text_delta)
- Handle StreamEvent to emit stream_delta type updates

Orchestrator changes:
- Feed tool calls and reasoning via append_tool(), response text via append_text() (stream_delta only, to avoid duplication)
- Skip progress_msg.edit_text when draft streaming is active

New settings: ENABLE_STREAM_DRAFTS, STREAM_DRAFT_INTERVAL

27 tests covering accumulation, throttle, composition, truncation, tool lines, and self-disable behavior.

### DIFF
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -710,7 +710,7 @@ class MessageOrchestrator:
                 for tc in update_obj.tool_calls:
                     name = tc.get("name", "unknown")
                     detail = self._summarize_tool_input(name, tc.get("input", {}))
-                    if not draft_streamer and verbose_level >= 1:
+                    if verbose_level >= 1:
                         tool_log.append({"kind": "tool", "name": name, "detail": detail})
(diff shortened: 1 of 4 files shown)


RichardAtCT:
Merged manually after rebasing to resolve conflicts with voice transcription PR. Included in main.

PR #124, opened 2026-03-04
docs: add Linux aiolimiter DBus workaround to setup guide

haripatel07:
## Description
Documents a known Linux-specific issue where Poetry fails to install 
`aiolimiter` due to a DBus/keyring error during `make dev`. The error 
is non-fatal so the install appears to succeed, but `aiolimiter` is 
missing at runtime — causing 2 test failures that are confusing to debug.

## Related Issue
N/A — discovered while setting up the dev environment on Linux (Ubuntu 24).

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [x] Documentation update

## Testing
- [x] Ran `make test` before fix: 2 failures
- [x] Ran `poetry run pip install aiolimiter` 
- [x] Ran `make test` after fix: 433 passed, 0 failures

## Checklist
- [x] Documentation updated
- [x] No breaking changes
- [x] Self-review completed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,13 +33,22 @@ This project is currently under active development with the following status:
    make dev
    ```
    > **Linux users**: If `make dev` shows a `DBusErrorResponse` / `ItemNotFoundException`
-   > error for `aiolimiter`, this is a known Poetry keyring issue on Linux and is
-   > non-fatal — the install continues. However, `aiolimiter` may not be installed,
-   > which will cause 2 test failures. Fix it by running:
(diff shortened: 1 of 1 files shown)


haripatel07:
Thanks for the thorough review! Regarding the poetry run pip install aiolimiter vs poetry add suggestion — aiolimiter is already declared as a dependency in pyproject.toml (via the python-telegram-bot[rate-limiter] extra). The issue is that Poetry silently skips installing it due to the DBus error, so poetry add would just update pyproject.toml unnecessarily. The pip install workaround is intentionally temporary — it just forces the already-declared package to actually land in the venv. Happy to add a clarifying comment in the docs if that would help.

PR #139, opened 2026-03-07
feat: add /model command for runtime model switching via Telegram

### LINKED ISSUE #138: feat: runtime model switching via /model command
**Is your feature request related to a problem?**
The Claude model can only be set at startup via `CLAUDE_MODEL` config. Users should be able to easily switch models on the go, similar to the Claude TUI interface.

**Describe the solution you'd like**
A `/model` command that allows per-user runtime model switching:
- `/model` — show current model
- `/model <name>` — switch model (any valid Claude model name)
- `/model default` — reset to server/CLI default

**Describe alternatives you've considered**
Changing the `CLAUDE_MODEL` env var and restarting the bot — requires server access, causes downtime, and affects all users.

**Additional context**
The override is per-user (stored in `context.user_data`) and does not affect other users. Implementation threads `model_override` through the facade and SDK integration layers.

bOsowski:
## Summary
- Add `/model` command for per-user runtime model switching
- Thread `model_override` through orchestrator → facade → SDK integration
- Extract actual model from API response (`AssistantMessage.model`)
- 14 new tests covering command behavior, parameter passthrough, and model override priority

## Related Issue
Fixes #138

## Type of Change
- [x] New feature

## Testing
- [x] Tests added/updated
- [x] All tests pass
- [x] Manual testing completed

## Checklist
- [x] Self-review completed
- [x] No breaking changes

### DIFF
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -626,6 +626,7 @@ class MessageOrchestrator:
         if not model_name or len(model_name) > 100:
             await update.message.reply_text("Invalid model name.")
             return
+        audit_logger = context.bot_data.get("audit_logger")
         if model_name == "default":
             context.user_data.pop("model_override", None)
             default = self._resolve_model_display(None, self.settings.claude_model)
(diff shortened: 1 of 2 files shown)


bOsowski:
> **Issues / questions**
> 
> 1. **[Important]** `src/bot/orchestrator.py` — No validation that the model name is a valid Claude identifier. A user can set `model_override` to any arbitrary string (up to 100 chars — e.g. `gpt-4`, `lol`, an empty string after `.strip()`). The SDK will fail at runtime with a potentially cryptic error rather than at set-time, leaving the user confused about why messages stop working. Consider either: (a) maintaining a loose allowlist of known `claude-*` prefixes and warning on mismatch, or (b) at minimum responding with a clear error if the next request fails due to an invalid model. Silent misconfiguration is poor UX.
F7C7D425-0264-43FC-9448-E5548538A03A
29EA4FB2-636F-4152-9024-EE1301E29DD6

This is handled - the response when an invalid model is selected is clear.


> 2. **[Important]** `src/bot/orchestrator.py` — `model_name` can technically be the empty string after `.strip()` if the input is `/model ` (one space). The length check `len(model_name) > 100` doesn't catch this — the existing `not model_name` check does, but only after `.strip()`. This is fine as-is, but worth confirming that the `not model_name` guard reliably covers the post-strip empty case. It does — just noting it so the test for this path is explicit. There's a test for `len > 100` but I don't see one for the whitespace-only input (`"/model   "`).

Telegram trims the command, such that the "/model    " becomes just "/model". This is reflected in the UI and it will trigger the base "/model" command.

> 3. **[Important]** `src/bot/orchestrator.py` — Audit log is placed outside the `if/else` block, so `/model default` (the reset path) logs `args=["default"]` rather than something semantically distinguishable from setting a model literally named "default". This makes audit log queries for "who changed their model to X" noisier. Suggest logging `args=["<reset>"]` or `args=[]` on the reset path, or restructuring the audit call to be inside each branch.

Will address in the next PR.

RichardAtCT:
Closing in favor of #160, which implements the same /model command with a more polished inline keyboard UI and per-model effort level support. Thank you for the initial implementation — it helped validate the feature design.

PR #140, opened 2026-03-07
fix: save image to temp file and include path in prompt for Claude CLI

### LINKED ISSUE #137: Image/screenshot upload with analysis doesn't work
Try uploading any image, it won't work :) 
Image

haripatel07:
## Problem
Image uploads were silently broken. `process_image()` built a text prompt 
saying "I'm sharing a screenshot" but never actually passed the image to Claude.
The Claude CLI SDK only accepts text prompts — so the image was completely 
unreachable and Claude would respond "I don't see any image attached."

## Root Cause
`ImageHandler.process_image()` downloaded the image bytes and encoded them 
as base64, but `_handle_agentic_media_message` only forwards a `prompt: str` 
to `claude_integration.run_command()`. The image data was stored in 
`ProcessedImage.base64_data` but never used.

## Fix
Download the image to `/tmp/claude_bot_files/image_<uuid>.<ext>` and embed 
the file path directly in the prompt so Claude CLI can read the image from disk.
```python
# Before — Claude never saw the image
prompt = "I'm sharing a screenshot with you. Please analyze it..."

# After — Claude can read the actual file
prompt = "I'm sharing a screenshot with you. The image is saved at: /tmp/claude_bot_files/image_abc123.png\n\nPlease analyze it..."
```

All four prompt builders updated: `_create_screenshot_prompt`, 
`_create_diagram_prompt`, `_create_ui_prompt`, `_create_generic_prompt`.

## Testing
- [x] 491 passed, 0 failed
- [x] No regressions

Closes #137

### DIFF
--- a/src/bot/features/image_handler.py
+++ b/src/bot/features/image_handler.py
@@ -55,8 +55,12 @@ class ImageHandler:
             )
         _TEMP_DIR.mkdir(mode=0o700, exist_ok=True)
         image_path = _TEMP_DIR / f"image_{uuid.uuid4()}.{fmt}"
-        image_path.write_bytes(image_bytes)
-        image_path.chmod(0o600)
+        try:
+            image_path.write_bytes(image_bytes)
+            image_path.chmod(0o600)
+        except Exception:
+            image_path.unlink(missing_ok=True)
+            raise
 
         # Detect image type for prompt tailoring
         image_type = self._detect_image_type(image_bytes)


RichardAtCT (commented):
**PR Review**
Reviewed head: `b00d01812c1ff0761b3f8d8b40117d027d6546cc`

**Summary**
- Fixes silently broken image uploads by saving the image to `/tmp/claude_bot_files/` and embedding the file path in the prompt, since the Claude CLI SDK only accepts text prompts
- Correct diagnosis of the root cause; approach is pragmatic given the SDK constraint

**What looks good**
- Root cause analysis is solid and the fix is the right shape given Claude CLI's text-only prompt interface
- `exist_ok=True` on `mkdir` avoids race conditions; `/tmp` always exists so no `parents=True` needed
- `fmt` is reused for both the extension and the metadata, avoiding a redundant second `_detect_format()` call
- 491 tests passing with no regressions

**Issues / questions**

1) **[Blocker]** `src/bot/features/image_handler.py` — **Does embedding a file path in the prompt actually cause Claude CLI to read the image?** The PR asserts this works but provides no evidence. If Claude CLI doesn't auto-read paths mentioned in prompts (it likely doesn't unless explicitly invoked with a file argument), the fix resolves nothing — it just changes the error from "I don't see an image" to "I can't read that path." Needs a smoke-test screenshot or confirmation of which Claude CLI flag/behaviour makes this work before merging.

2) **[Important]** `src/bot/features/image_handler.py` — **No temp file cleanup.** Every image processed writes a file that is never deleted. On a busy bot this is an unbounded disk leak. Add a `try/finally` that unlinks the path after `run_command()` completes, or use `tempfile.NamedTemporaryFile(delete=True)` scoped to the handler call.

3) **[Important]** `src/bot/features/image_handler.py` — **`import uuid` is inside the function body.** CLAUDE.md explicitly requires module-level imports. Move it to the top of the file next to the other stdlib imports.

4) **[Important]** `src/bot/features/image_handler.py` — **`ProcessedImage.base64_data` is now dead code.** The base64 is still computed and stored but nothing uses it (the comment `# Convert to base64 for Claude (if supported in future)` was removed). Either remove the field + computation to avoid confusion, or leave the comment explaining why it's retained. As-is it's misleading.

5) **[Nit]** `src/bot/features/image_handler.py` — `_detect_image_type()` is now called with `bytes(image_bytes)` after `_detect_format()` already converted it. The double conversion is harmless but consider assigning `image_bytes_bytes = bytes(image_bytes)` once at the top and reusing it.

**Suggested tests (if needed)**
- Integration test: upload an image, assert the generated prompt string contains a valid path under `/tmp/claude_bot_files/`, and assert that path exists on disk at the time `run_command()` is called
- Unit test: assert temp file is cleaned up after processing completes (once cleanup is added)
- Confirm (manually or in CI) that the Claude CLI actually reads the image when given a path in the prompt text — this is the core assumption the fix rests on

**Verdict**
⚠️ Merge after fixes — the disk-leak and dead-code issues need addressing, and the `import uuid` is a quick style fix. More critically, the fundamental assumption (Claude CLI reads image paths from prompt text) should be validated before this ships; if it doesn't work that way, the approach needs rethinking entirely.

— *Friday, AI assistant to @RichardAtCT*

haripatel07:
Addressed all review feedback in 62f27f7:

- `uuid` moved to module scope
- Format validated before writing to disk — `ValueError` raised for unknown formats, `.jpg` fallback removed
- `0o700` dir + `0o600` file permissions
- Single `bytes()` conversion at top of method
- All four prompt builders now explicitly say `"Please read the image file at this path using your Read tool: {path}"` — this is how Claude CLI accesses the image since the SDK has no native image attachment API (confirmed by inspecting `client.py`)
- `finally` block in `agentic_photo` deletes temp file after `run_command` completes
- `base64_data` retained with comment explaining it's reserved for future multimodal SDK support

491 passing, 0 failures.

haripatel07:
Fixed in 48612f2:

- Wrapped `write_bytes` + `chmod` in try/except — partial file unlinked on failure before re-raising
- Type hints added to all four `_create_*_prompt` methods (CLAUDE.md compliance)

491 passing.

RichardAtCT:
Closing in favor of #168, which sends images as multimodal SDK content blocks (base64 + media type) rather than file paths in prompts. The SDK multimodal approach is the proper integration path for Claude. Thank you for diagnosing the root cause of the image upload issue — your analysis was valuable.

PR #147, opened 2026-03-15
feat: follow-up message interrupt (like vanilla Claude Code)

Fr4nzz:
## Summary
- When a user sends a follow-up message while Claude is processing, the bot interrupts the current task and immediately processes the follow-up in the same session
- Uses the SDK's `client.interrupt()` control protocol (equivalent to Ctrl+C)
- If graceful interrupt doesn't stop within 3 seconds, falls back to killing the subprocess
- Session history is preserved across interrupts — Claude remembers everything
- Adds `/stop` command for manually cancelling tasks
- Enables `concurrent_updates(True)` so follow-up messages are handled in parallel

## How it works
1. User sends message A → Claude starts processing
2. User sends message B while Claude is working
3. Bot sends `interrupt()` signal to Claude (like Ctrl+C)
4. Claude stops, session ID is preserved
5. Bot immediately sends message B as a continuation of the same session
6. Claude sees full conversation history + message B

## Test plan
- [ ] Send a long task, then send a follow-up while processing
- [ ] Verify `📨 Interrupting...` appears and the follow-up is processed
- [ ] Verify `/stop` cancels a running task
- [ ] Verify session continues after interrupt (Claude remembers context)
- [ ] Verify normal sequential messages still work when no task is running

### DIFF
--- a/src/bot/core.py
+++ b/src/bot/core.py
@@ -53,6 +53,7 @@ class ClaudeCodeBot:
         builder.token(self.settings.telegram_token_str)
         builder.defaults(Defaults(do_quote=self.settings.reply_quote))
         builder.rate_limiter(AIORateLimiter(max_retries=1))
+        builder.concurrent_updates(True)
 
         # Configure connection settings
         builder.connect_timeout(30)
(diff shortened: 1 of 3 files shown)


RichardAtCT:
Closing in favor of #152, which bundles this follow-up interrupt feature together with streaming drafts, enhanced UX, and rich formatting.

PR #148, opened 2026-03-15
feat: enhanced UX with per-event messages and file handling

Fr4nzz:
## Summary
- **Per-event tool messages**: Each tool call (Read, Bash, Grep, etc.) gets its own message with icon, instead of cramming everything into one progress message
- **Extended thinking**: Show Claude's ThinkingBlocks as 🧠 messages
- **Intermediate reasoning**: Show visible text between tool calls as 💬 messages
- **`/cleanup` command**: Delete tool/thinking messages from the last response on demand
- **Progress timer**: `⏳ Working... (Xs)` updates every 3 seconds, finalizes as `✅ Done (Xs)`
- **File handling**: Rename `send_image_to_user` → `send_file_to_user` to accept any file type
- **Upload handling**: Download files to disk and pass path to Claude (instead of base64)
- **Security**: Skip file extension whitelist when `DISABLE_SECURITY_PATTERNS=true`
- **HTML fallback**: Strip tags in plain text fallback instead of showing raw `<b>` tags
- **`_send_formatted_message` helper**: Consistent HTML-with-fallback across all handlers

## Test plan
- [ ] Send a text message and verify tool calls appear as individual messages
- [ ] Verify 🧠 and 💬 messages appear during processing
- [ ] Use `/cleanup` to delete tool messages after response
- [ ] Verify progress shows elapsed time and finalizes as ✅ Done
- [ ] Upload a file and verify Claude can read it
- [ ] Send an image and verify Claude receives the file path

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #146, opened 2026-03-15
feat: streaming draft responses via sendMessageDraft

Fr4nzz:
## Summary
- Rewrite `DraftStreamer` to support both `sendMessageDraft` (private chats) and `editMessageText` (group chat fallback)
- Add cursor `▌` during streaming for visual typing feedback (like OpenClaw)
- Minimum 20 chars before first draft send (avoids triggering push notifications with fragments)
- Anti-regressive updates: skip edits where text got shorter
- Smart error classification: distinguish draft-unavailable, message-not-modified, message-deleted
- Self-disabling after 3 consecutive errors
- Add `ENABLE_STREAM_DRAFTS` and `STREAM_DRAFT_INTERVAL` config options
- Support verbose level 3 (full command output)

## Test plan
- [ ] Enable `ENABLE_STREAM_DRAFTS=true` and send a text prompt
- [ ] Verify response text streams token-by-token with cursor ▌
- [ ] Verify cursor disappears when response is finalized
- [ ] Test in group chat: should fall back to editMessageText
- [ ] Verify no draft sent before 20 characters accumulated

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #145, opened 2026-03-15
feat: rich HTML formatting for Telegram messages

Fr4nzz:
## Summary
- Convert markdown tables to aligned monospace `<pre>` blocks with box-drawing characters
- Add native `<blockquote>` support for markdown blockquotes
- Add bullet list (• ), numbered list, and horizontal rule conversion
- Fix HTML tag nesting issues (`<b><i>...</i></b>` vs `<i><b>...</i></b>`) that cause Telegram to reject messages
- Add `_repair_html_nesting()` that automatically fixes overlapping inline tags

## Test plan
- [ ] Send Claude a prompt asking for tables, code blocks, blockquotes, lists
- [ ] Verify tables render in monospace `<pre>` blocks with aligned columns
- [ ] Verify blockquotes use native Telegram `<blockquote>` tags
- [ ] Verify bold+italic nesting doesn't cause "Can't parse entities" errors
- [ ] Test fallback: malformed HTML stripped to plain text instead of showing raw tags

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #126, opened 2026-03-04
Feature: Allow sending follow-up messages during active Claude task


vckobi:
### Problem
When Claude is actively working on a task (reading files, editing code, running commands), the user cannot send additional context, corrections, or notes until the current operation completes.

Claude Code CLI natively supports this — you can type messages while Claude is working and it picks them up. However, the Telegram bot currently blocks input during active tasks, requiring fully sequential interaction.

### Proposed Solution
Allow users to send messages while Claude is processing a request. These messages could either:

1. **Queue & append** — Collect follow-up messages and feed them to Claude as additional context when the current turn ends (simplest approach)
2. **Live inject** — Pass the message to the active Claude session in real-time, similar to how the CLI works (ideal but more complex)
3. **Hybrid** — Let short messages queue, but support a prefix like `!` or `/note` to inject immediately

### Use Cases
- User realizes they forgot to mention a constraint → sends "also make sure to handle null values"
- User wants to redirect Claude mid-task → "actually, use TypeScript instead"
- User wants to add context → "the config file is at /etc/app/config.yaml"
- User spots Claude going in wrong direction → "stop, that's the wrong file"

### Related PRs
- #116 (`/stop` command) — covers cancellation but not appending context
- #122 (inline Stop button) — same, cancel only
- #123 (streaming responses) — shows progress but no input during task
- #102 (concurrent processing across topics) — parallel topics, not mid-task input

### Environment
Using claude-code-telegram v1.3.0 inside a Docker container (linuxserver/code-server) with Claude Code CLI.

xiaotonng:
Standard stdin blocking makes this tricky. In [pikiclaw](https://github.com/xiaotonng/pikiclaw), we solved this by decoupling the IM event queue from the stream loop via an async watchdog, allowing interruptions or follow-ups via Telegram mid-generation. Feel free to reference our approach or test it via `npx pikiclaw@latest`.

RichardAtCT:
Follow-up messages during active tasks are tracked by PR #152 (bundled UX improvements, includes follow-up interrupt support). The interrupt approach in #152 allows users to send messages that redirect Claude mid-task, similar to the vanilla Claude Code CLI behavior. Will merge once #152 is rebased on main.

PR #152, opened 2026-03-15
feat: streaming drafts, rich formatting, follow-up interrupts, and UX improvements

Fr4nzz:
## Summary

- **Streaming draft responses** via `sendMessageDraft` — real-time token streaming in private chats with fallback to `editMessageText` for groups. Draft cleared before final response to prevent stale bubbles; draft text reset after each 💬 message to avoid accumulation
- **Rich HTML formatting** — parse Claude's markdown into Telegram HTML (bold, italic, code, links) with smart message splitting
- **Follow-up message interrupts** — send a new message while Claude is working to redirect, like vanilla Claude Code
- **Per-event verbose messages** — tool calls, thinking, and assistant text shown as individual messages with configurable verbosity (0/1/2)
- **Image sending via MCP** — intercept `send_file_to_user` tool calls, validate paths, send as Telegram photos
- **SDK integration** — use `setting_sources=["project"]` to avoid plugin MCP conflicts, support SSE-based MCP servers
- **Keep progress messages visible** after response with elapsed time

## Test plan

- [ ] Send a text message and verify streaming draft appears in real-time
- [ ] Send a follow-up message while Claude is working — verify it interrupts
- [ ] Test `/verbose 0`, `/verbose 1`, `/verbose 2` — verify output levels
- [ ] Test image sending via Playwright screenshot + MCP `send_file_to_user`
- [ ] Verify draft bubble clears before final response appears
- [ ] Test in group chat (editMessageText fallback)
- [ ] Run `pytest tests/unit/ -o "addopts="` — all tests pass

### DIFF
--- a/src/bot/core.py
+++ b/src/bot/core.py
@@ -53,6 +53,9 @@ class ClaudeCodeBot:
         builder.token(self.settings.telegram_token_str)
         builder.defaults(Defaults(do_quote=self.settings.reply_quote))
         builder.rate_limiter(AIORateLimiter(max_retries=1))
+        # Allow concurrent update processing so follow-up messages
+        # can interrupt a running Claude task
+        builder.concurrent_updates(True)
 
         # Configure connection settings
         builder.connect_timeout(30)
--- a/src/bot/handlers/message.py
+++ b/src/bot/handlers/message.py
@@ -360,13 +360,13 @@ async def handle_text_message(
 
         # Enhanced stream updates handler with progress tracking
         async def stream_handler(update_obj):
-            # Intercept send_image_to_user MCP tool calls.
+            # Intercept send_file_to_user / send_image_to_user MCP tool calls.
             # The SDK namespaces MCP tools as "mcp__<server>__<tool>".
             if update_obj.tool_calls:
                 for tc in update_obj.tool_calls:
                     tc_name = tc.get("name", "")
-                    if tc_name == "send_image_to_user" or tc_name.endswith(
-                        "__send_image_to_user"
+                    if tc_name in ("send_file_to_user", "send_image_to_user") or tc_name.endswith(
+                        ("__send_file_to_user", "__send_image_to_user")
                     ):
                         tc_input = tc.get("input", {})
                         file_path = tc_input.get("file_path", "")
@@ -439,7 +439,7 @@ async def handle_text_message(
         # Delete progress message
         await progress_msg.delete()
 
-        # Use MCP-collected images (from send_image_to_user tool calls)
+        # Use MCP-collected files (from send_file_to_user tool calls)
         images: list[ImageAttachment] = mcp_images
 
         # Try to combine text + images when response fits in a caption
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -34,8 +34,10 @@ from ..projects import PrivateTopicsUnavailableError
 from .utils.draft_streamer import DraftStreamer, generate_draft_id
 from .utils.html_format import escape_html
 from .utils.image_extractor import (
+    FileAttachment,
     ImageAttachment,
     should_send_as_photo,
+    validate_file_path,
     validate_image_path,
 )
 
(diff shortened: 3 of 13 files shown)


RichardAtCT:
Big feature PR with some genuinely useful additions — streaming drafts and interrupt support are quality-of-life wins. Several correctness issues and a process concern before this can land.

---

### 🔴 Blockers

**1. `/stop` is a no-op — `running_claude_task` is never set**

`agentic_stop` reads `context.user_data["running_claude_task"]` to cancel the task, but nowhere in the diff is that key actually written. The command will always hit the `else` branch and say "Nothing running." You need something like:

```python
task = asyncio.create_task(self._run_agentic(...))
context.user_data["running_claude_task"] = task
```

before any `await` in the message handler. Without this, the entire interrupt mechanism is non-functional.

**2. `concurrent_updates(True)` makes the race worse without the fix above**

Enabling concurrent updates means multiple messages process simultaneously — right intent for interrupt support. But if `/stop` can't actually cancel the running task, you now have concurrent handlers stepping on each other with no abort path. Don't ship `concurrent_updates(True)` until the task lifecycle is wired up correctly.

---

### 🟡 Consistency / Design Issues

**3. `user_data` vs `chat_data` — conflicts with PR #165**

`agentic_cleanup` reads `last_tool_message_ids` and `last_tool_chat_id` from `context.user_data`. PR #165 (also open, also touches `orchestrator.py`) moved session state to `chat_data`. These two PRs will conflict on merge and produce inconsistent state. Agree on `chat_data` as the canonical store before either lands.

**4. `tool_log` used as a mixed-type container**

```python
tool_log.append({"_tool_message_ids": tool_message_ids})
```

Injecting a sentinel dict into a list of tool call entries is a code smell — any downstream iterator has to defensively check for this magic key. Track `tool_message_ids` as a separate variable in the closure scope instead.

---

### 🟡 Documentation Drift

**5. Verbose level 3 undocumented / CLAUDE.md out of date**

CLAUDE.md config docs apparently still document verbose range as 0–2. Update to reflect level 3 and describe what it enables.

---

### 🟡 Process

**6. Overlap with PR #148 (same author)**

PR #148 is already open with very similar `orchestrator.py` changes from the same author. These will be nearly impossible to merge independently. Recommend either:
- Explicitly closing #148 and noting this PR supersedes it, or
- Rebasing this on top of #148 after that merges

As-is, whichever merges second will have significant conflicts.

---

### 🟢 Test gaps

No tests visible for:
- `agentic_stop` / `agentic_cleanup` handlers
- `draft_streamer.py` (private chat vs group fallback)
- `html_format.py` markdown→HTML edge cases (nested formatting, split threshold)

---

### Minor

- `except Exception: pass` in `_send_tool_msg` and `agentic_cleanup` silently eats errors — log at WARNING with structlog
- `agentic_stop` and `agentic_cleanup` missing type hints on `update`/`context` — mypy strict violation

---

**Summary:** Streaming draft and HTML formatting look solid in concept, but `/stop` is broken as written (task never stored), and `concurrent_updates(True)` actively makes things worse without it. Fix the task lifecycle, resolve the `user_data`/`chat_data` conflict with #165, and sort out the #148 overlap before merging.

RichardAtCT:
This is the authoritative bundle for streaming drafts, rich formatting, follow-up interrupts, and UX improvements. We're closing #145, #146, #147, and #148 in favor of this PR.

**Action needed**: The merge state is currently DIRTY (conflicts with main). Please rebase on main — we're merging a batch of 7 PRs that will change the baseline. After rebase:
1. Ensure all tests pass
2. Verify no regressions with the streaming drafts feature from #123 (already in main)

The feature set looks comprehensive. Looking forward to merging once conflicts are resolved.

RichardAtCT:
Thanks, and sorry this sat so long.

Some of it has since landed on main from other PRs: streaming drafts in #123 and the Stop button in #122. So this no longer applies cleanly — it conflicts and sits 31 commits behind.

It is also four features in one branch with 485 deletions and a 643-line rewrite of `orchestrator.py`. I cannot review that as a unit.

If you still want the follow-up interrupt work, open it on its own against current main. It maps to open issue #126 and nobody is on it.

PR #158, opened 2026-03-20
Add make run-watch for auto-restart during development

thereisnotime:
## Summary

- Adds a `make run-watch` target that uses `watchfiles` to auto-restart the bot when files in `src/` change
- `watchfiles` is already an indirect dependency (pulled in by `uvicorn`), so no new deps needed
- Updated README Development section to mention the new target

## Test plan

- [ ] Run `make run-watch`, edit a file in `src/`, confirm the bot restarts automatically
- [ ] Run `make help`, confirm `run-watch` shows up in the list

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -140,6 +140,34 @@ QUICK_ACTIONS_TIMEOUT=120
 # Git operations timeout in seconds
 GIT_OPERATIONS_TIMEOUT=30
 
+# === VOICE TRANSCRIPTION ===
+# Enable voice message transcription
+ENABLE_VOICE_MESSAGES=true
+
+# Voice transcription provider: mistral, openai, or local
+# - mistral: Uses Mistral Voxtral (requires MISTRAL_API_KEY)
+# - openai: Uses OpenAI Whisper API (requires OPENAI_API_KEY)
+# - local: Uses whisper.cpp binary (requires ffmpeg + whisper.cpp installed)
+VOICE_PROVIDER=mistral
+
+# API keys (only needed for cloud providers)
+MISTRAL_API_KEY=
+OPENAI_API_KEY=
+
+# Override transcription model (optional)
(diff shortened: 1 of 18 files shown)


RichardAtCT:
Good DX addition. A few things worth considering:

**1. Invocation is correct** — `watchfiles "command" path` is the right CLI syntax. No issues there.

**2. Indirect dep is the real risk here.** `watchfiles` is pulled in by `uvicorn`, but if a user isn't running the API server (or uvicorn gets dropped/replaced in a future refactor), `poetry run watchfiles` will silently fail with a confusing "command not found" error. Since this is a dev convenience target, it's worth either:
- Adding `watchfiles` as an explicit dev dependency: `poetry add --group dev watchfiles`, or
- Adding a guard comment making the implicit dep explicit (at minimum)

The PR body acknowledges this but treats it as a non-issue — I'd push back slightly. Indirect deps are fragile anchors.

**3. Minor: `python -m src.main` vs the entrypoint.** The `run` target uses `poetry run claude-telegram-bot` (the installed entrypoint). `run-watch` uses `python -m src.main` directly. These may behave identically, but if the entrypoint ever does setup work (env loading, etc.), they'll diverge silently. Prefer consistency — either use the entrypoint via `watchfiles "claude-telegram-bot" src/`, or document why the module invocation is intentional.

**4. Signal handling** — `watchfiles` sends `SIGTERM` on restart by default, which is fine for most cases. No action needed unless the bot has a custom signal handler.

Overall: clean PR, just pin `watchfiles` as an explicit dev dep to avoid a future footgun.

thereisnotime:
Hey @RichardAtCT — both points addressed:

1. **Explicit dev dep** — `watchfiles` is now a proper dev dependency instead of relying on uvicorn pulling it in
2. **Consistent entrypoint** — `run-watch` now uses `claude-telegram-bot` matching the `run` target

Ready for another look!

Issue #121, opened 2026-03-01
Sonnet by default


chrisgwynne:
Whenever i start a new session it starts with Opus. I ask for Sonnet always. But that doesn't stick. Any .env setting?

chrisgwynne:
Root cause found. In sdk_integration.py:197, ClaudeAgentOptions is built without model=self.config.claude_model,
  so it defaults to None — meaning no --model flag is passed to the CLI, which then uses its default (Opus 4.6).    
                                                                                                                    
```
Update(.local/share/uv/tools/claude-code-telegram/lib/python3.14/site-packages/src/claude/sdk_integration.py)
  ⎿  Added 1 line                                                                                                   
      201                  allowed_tools=sdk_allowed_tools,                                                         
      202                  disallowed_tools=sdk_disallowed_tools,
      203                  cli_path=self.config.claude_cli_path or None,                                            
      204 +                model=self.config.claude_model or None,                                           
      205                  sandbox={                                                                                
      206                      "enabled": self.config.sandbox_enabled,
      207                      "autoAllowBashIfSandboxed": True,

● Now add CLAUDE_MODEL to .env so it's explicit (and points to current Sonnet):

● Update(.config/claude-telegram/.env)                                                                              
  ⎿  Added 1 line
      3                                                                           
      4  
      5                                                                                                             
      6 +CLAUDE_MODEL=claude-sonnet-4-6                                                                      
      7  
      8  
      9  
```
  Bug: sdk_integration.py built ClaudeAgentOptions without the model parameter, so it defaulted to None → no --model
   flag → Claude CLI used its default (Opus 4.6).
                                                                                                                    
  Fix:            
  1. Added model=self.config.claude_model or None to ClaudeAgentOptions in sdk_integration.py:204
  2. Added CLAUDE_MODEL=claude-sonnet-4-6 to .env to explicitly set the current Sonnet model (the code default was
  the older claude-3-5-sonnet-20241022)
  3. Restarted the service

  New Telegram sessions should now use Sonnet 4.6.

PR #160, opened 2026-03-20
feat: runtime model switching via /model command

### LINKED ISSUE #138: feat: runtime model switching via /model command

Talla:
## Summary

Adds a `/model` command that lets users switch between Claude models (Opus, Sonnet, Haiku) and set effort levels at runtime via an inline keyboard UI — no bot restart or config changes needed.

- **Inline keyboard flow**: `/model` → pick model → pick effort level (if supported) → done
- **Model-aware effort levels**: Haiku has no effort support, Sonnet supports low/medium/high, Opus adds "max"
- **Session handling**: Switching models forces a new session (CLI doesn't support model changes on resumed sessions). Users are warned upfront before selecting.
- **Per-user override**: Stored in `context.user_data` (in-memory). Resets on bot restart — server default (`CLAUDE_MODEL` env var) is used until the user overrides.

Implements the feature requested in #138.

## Changes

| File | What |
|------|------|
| `command.py` | `/model` command, `model_callback`, `_handle_model_selection` shared logic, model/effort constants |
| `callback.py` | Route `model:` and `effort:` callbacks in classic mode |
| `orchestrator.py` | Register command + callback in both modes, add to bot menu, thread overrides to `run_command` |
| `message.py` | Thread `model_override` + `effort_override` through all 4 classic `run_command` call sites |
| `facade.py` | Accept and forward `model_override` + `effort_override` params |
| `sdk_integration.py` | Apply `model_override` and `effort` to `ClaudeAgentOptions` |
| `test_model_command.py` | 17 new tests (keyboard, selection, labels, effort config) |
| `test_orchestrator.py` | Updated handler count assertions |

## Test plan

- [x] All 529 tests pass (512 existing + 17 new)
- [x] Manual: `/model` shows keyboard with Opus/Sonnet/Haiku + Reset
- [x] Manual: Selecting Haiku skips effort selection
- [x] Manual: Selecting Sonnet shows Low/Medium/High (no Max)
- [x] Manual: Selecting Opus shows Low/Medium/High/Max
- [x] Manual: After switching, next message uses the new model
- [x] Manual: "Reset to default" clears override

Closes #138

### DIFF
--- a/src/bot/handlers/callback.py
+++ b/src/bot/handlers/callback.py
@@ -57,6 +57,11 @@ async def handle_callback_query(
             action, param = data, None
 
         # Route to appropriate handler
+        from .command import _handle_model_selection
+
+        async def _model_effort_handler(query, param, context):
+            await _handle_model_selection(query, f"{action}:{param}", context)
+
         handlers = {
             "cd": handle_cd_callback,
             "action": handle_action_callback,
@@ -66,6 +71,8 @@ async def handle_callback_query(
             "conversation": handle_conversation_callback,
             "git": handle_git_callback,
             "export": handle_export_callback,
+            "model": _model_effort_handler,
+            "effort": _model_effort_handler,
         }
 
         handler = handlers.get(action)
--- a/src/bot/handlers/callback.py
+++ b/src/bot/handlers/callback.py
@@ -12,6 +12,7 @@ from ...config.settings import Settings
 from ...security.audit import AuditLogger
 from ...security.validators import SecurityValidator
 from ..utils.html_format import escape_html
+from .command import _handle_model_selection
 
 logger = structlog.get_logger()
 
@@ -57,11 +58,6 @@ async def handle_callback_query(
             action, param = data, None
 
         # Route to appropriate handler
-        from .command import _handle_model_selection
-
-        async def _model_effort_handler(query, param, context):
-            await _handle_model_selection(query, f"{action}:{param}", context)
-
         handlers = {
             "cd": handle_cd_callback,
             "action": handle_action_callback,
@@ -71,8 +67,8 @@ async def handle_callback_query(
             "conversation": handle_conversation_callback,
             "git": handle_git_callback,
             "export": handle_export_callback,
-            "model": _model_effort_handler,
-            "effort": _model_effort_handler,
+            "model": lambda q, p, ctx: _handle_model_selection(q, f"model:{p}", ctx),
+            "effort": lambda q, p, ctx: _handle_model_selection(q, f"effort:{p}", ctx),
         }
 
         handler = handlers.get(action)
--- a/src/bot/handlers/command.py
+++ b/src/bot/handlers/command.py
@@ -174,7 +174,8 @@ async def help_command(update: Update, context: ContextTypes.DEFAULT_TYPE) -> No
         "• <code>/status</code> - Show session and usage status\n"
         "• <code>/export</code> - Export session history\n"
         "• <code>/actions</code> - Show context-aware quick actions\n"
-        "• <code>/git</code> - Git repository information\n\n"
+        "• <code>/git</code> - Git repository information\n"
+        "• <code>/model [name]</code> - View or switch Claude model\n\n"
         "<b>Session Behavior:</b>\n"
         "• Sessions are automatically maintained per project directory\n"
         "• Switching directories with <code>/cd</code> resumes the session for that project\n"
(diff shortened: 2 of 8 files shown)


RichardAtCT:
Good implementation overall — the UX flow is clean and the test coverage is solid. Six issues to address before merging:

---

**🔴 Bug: `action` closure in `_model_effort_handler` (callback.py)**

This is the most critical issue. `_model_effort_handler` references `action` from the outer scope, but `action` is a loop variable (or however `handlers` is constructed). Both `"model"` and `"effort"` entries point to the same closure, and `action` will be whatever value it held last when the dict was defined — likely `"effort"` for both handlers. This means a `model:opus` callback will be reconstructed as `effort:opus` and silently fail.

Fix by capturing explicitly:

```python
async def _make_model_effort_handler(action: str):
    async def handler(query, param, context) -> None:
        await _handle_model_selection(query, f"{action}:{param}", context)
    return handler
```

Or just inline two named functions — cleaner and avoids the factory pattern for async.

---

**🔴 Missing type annotations (mypy strict)**

`_handle_model_selection` is missing return type and parameter annotations. With `disallow_untyped_defs=true` this will fail CI. Should be:

```python
async def _handle_model_selection(
    query: CallbackQuery,
    data: str,
    context: ContextTypes.DEFAULT_TYPE,
) -> None:
```

Same applies to `model_command` — `update` and `context` need annotations.

---

**🟡 `effort_override` not visibly applied in `sdk_integration.py`**

The diff shows `model_override` wired into `ClaudeAgentOptions` but `effort_override` is elided. If it's genuinely there in the full diff, fine — but if not, this is a silent no-op: users select an effort level, get confirmation text, but the SDK ignores it. Needs explicit confirmation.

---

**🟡 Hardcoded model IDs will go stale**

`"haiku": "claude-haiku-4-5-20251001"` is already date-versioned and will need updating as Anthropic releases new versions. Consider moving `_MODELS` to config/env so it's overridable without a deploy, or at minimum add a comment linking to the Anthropic model docs.

---

**🟡 `user_data` vs `chat_data` — potential conflict with PR #165**

PR #165 migrates session state to `chat_data`. This PR stores model/effort overrides in `user_data`. In a group chat, one user's model choice would only affect their `user_data` while the session (in `chat_data`) is shared — meaning the model applied depends on who sent the last message. These two PRs need to be reconciled before either merges.

---

**🟢 Test quality — mostly good**

17 tests covering happy path, reset, Haiku (no effort), and effort selection is solid. Two suggestions:
- Add a test for the `action` closure bug (would have caught it)
- Add a test asserting `force_new_session` is set on model change

---

**Summary:** The `action` closure bug is a functional regression that will silently swallow model selections. The mypy annotations are a CI blocker under strict mode. Everything else is addressable before or alongside merge.

RichardAtCT:
This is the preferred /model implementation — closing #139 in its favor. The inline keyboard UI and per-model effort levels are a nice touch.

**Action needed**: The branch is currently in CONFLICTING state. Please rebase on main (we're merging several PRs that touch orchestrator.py). Once rebased with passing tests, this should be ready to merge.

Note: #139 has been closed as superseded by this PR.

Talla:
Thanks for the detailed review! All points addressed in this force-push. Here's what was done:

---

**🔴 Bug: `action` closure in `_model_effort_handler`**
Fixed. Replaced the shared closure with two explicit lambdas, each hardcoding its own prefix:
```python
"model": lambda q, p, ctx: _handle_model_selection(q, f"model:{p}", ctx),
"effort": lambda q, p, ctx: _handle_model_selection(q, f"effort:{p}", ctx),
```
Also moved the import to module level. Added a regression test that would have caught this.

**🔴 Missing type annotations**
Fixed. `_handle_model_selection` now has `query: CallbackQuery, data: str, context: ContextTypes.DEFAULT_TYPE`.

**🟡 `effort_override` not visibly applied in `sdk_integration.py`**
It was wired (`effort=effort_override` in `ClaudeAgentOptions`) — just not visible in the diff due to context. Added an explicit log line to make it obvious:
```python
if model_override or effort_override:
    logger.info("Runtime model/effort override active", model_override=..., effort_override=...)
```

**🟡 Hardcoded model IDs will go stale**
Went further than a comment. Dropped `_MODELS` entirely and replaced with short CLI aliases (`"opus"`, `"sonnet"`, `"haiku"`), which the Claude CLI resolves to the current latest of each family automatically — no version numbers to maintain anywhere. Confirmed these are accepted by the CLI (`--model` flag docs) and the SDK types (`Literal["sonnet", "opus", "haiku"]`).

**🟡 `user_data` vs `chat_data` — conflict with PR #165**
PR #165 is not merged yet and our PR should go in independently. `model_override` and `effort_override` intentionally stay in `user_data` (per-user model preference makes sense semantically). `force_new_session` also stays in `user_data` to be consistent with current `main`. Added a comment flagging the dependency:
```python
# Note: if PR #165 merges first, change this to context.chat_data
context.user_data["force_new_session"] = True
```
When #165 merges, that one-line change will be needed here too.

**Tests**
Added three new tests:
- Closure regression guard: asserts `model:sonnet` sets `model_override`, not `effort_override`
- Effort prefix isolation: asserts `effort:high` doesn't overwrite `model_override`
- `force_new_session` assertion on model switch

**Rebase**
Rebased onto current `main` (10 commits ahead). Conflicts resolved in `sdk_integration.py`, `facade.py`, `orchestrator.py`, and `test_orchestrator.py` — all tests pass (552 total).

Clawdy-ast:
After cherry-picking this PR onto v1.6.0, hit a regression: `.env CLAUDE_MODEL` is no longer honoured as the global default. The bot falls back to Sonnet (SDK default) even when `.env CLAUDE_MODEL=claude-opus-4-7` is set, as long as the user has no per-user override stored from `/model`.

**Repro:**
1. Cherry-pick this PR onto v1.6.0 (or run on `feat/model-command` directly)
2. Set `.env CLAUDE_MODEL=claude-opus-4-7`, leave per-user override DB empty
3. Restart bot
4. Send any message from a user that has never run `/model`
5. Inspect `ResultMessage.model_usage` (or `actual_models` field if logged) — shows Sonnet, not Opus

**Expected:** resolution chain should be `per-user override → .env CLAUDE_MODEL → SDK default`. With no override, `.env` should win.

**Looks like:** the per-user-override resolution path is short-circuiting to the SDK default when no override exists, bypassing the `.env` value that issue #121's fix wired into `ClaudeAgentOptions(model=...)`.

**Workaround:** set the model via `/model` once per user — works fine (that's what this PR adds). The `.env` value just stops mattering as a default.

**Env:** Windows 11, Python 3.14, latest `claude-agent-sdk`, bot v1.6.0 + this PR cherry-picked locally.

Could be a quick fix once the resolution function is found — happy to PR if pointed at the right module.

PR #162, opened 2026-03-22
Retrieving specific field for TextBlock and ThinkingBlock

### LINKED ISSUE #161: ThinkingBlock prints object to message instead of text
### Describe the bug
While streaming claude messages, ThinkingBlock gets printed like so:

> Working... (8s)
> 
> 💬 [ThinkingBlock(thinking='The user is asking "what is the status" - this is vague

### To Reproduce
Just any prompt for Claude to think.

### Environment
Project Version: 1.5.0

tolstovdmit:
## Description
Instead of attr check, checked type. Also changed getattr to field call Issue  

## Related Issue
Fixes #161

## Type of Change
- [x] Bug fix

## Testing
- [x] Manual testing completed

## Checklist
- [x] Self-review completed
- [ ] Documentation updated
- [x] No breaking changes (or clearly documented)

### DIFF
--- a/src/claude/sdk_integration.py
+++ b/src/claude/sdk_integration.py
@@ -20,6 +20,8 @@ from claude_agent_sdk import (
     PermissionResultDeny,
     ProcessError,
     ResultMessage,
+    TextBlock,
+    ThinkingBlock,
     ToolPermissionContext,
     ToolUseBlock,
     UserMessage,
@@ -496,13 +498,15 @@ class ClaudeSDKManager:
                         if isinstance(block, ToolUseBlock):
                             tool_calls.append(
                                 {
-                                    "name": getattr(block, "name", "unknown"),
-                                    "input": getattr(block, "input", {}),
-                                    "id": getattr(block, "id", None),
+                                    "name": block.name,
+                                    "input": block.input,
+                                    "id": block.id,
                                 }
                             )
-                        elif hasattr(block, "text"):
+                        elif isinstance(block, TextBlock):
                             text_parts.append(block.text)
+                        elif isinstance(block, ThinkingBlock):
+                            text_parts.append(block.thinking)
 
                 if text_parts or tool_calls:
                     update = StreamUpdate(


RichardAtCT:
Good fix overall — the `isinstance` approach is strictly better than `hasattr` for typed SDK objects, and the direct field access on `ToolUseBlock` is safe since these are typed dataclasses/Pydantic models where the fields are guaranteed present.

A few things worth considering:

**ThinkingBlock → text_parts: debatable**

Appending `block.thinking` to `text_parts` means thinking content gets surfaced to users alongside regular text. Depending on how `text_parts` is consumed downstream (e.g. sent directly to Telegram), this could expose raw chain-of-thought in user-facing messages. Is that intentional? A common pattern is to either discard thinking blocks or accumulate them separately (e.g. `thinking_parts`) and only log/debug them. Worth a comment in the code at minimum confirming this is deliberate.

**ToolUseBlock direct access**

The switch from `getattr(block, "name", "unknown")` to `block.name` is correct *if* the SDK guarantees these fields are always populated. If the SDK can return a partial/streaming `ToolUseBlock` with `name=None` mid-stream, you've traded a safe default for a potential `None` propagation. Quick sanity check: does the SDK type stub for `ToolUseBlock.name` show `str` or `Optional[str]`? If `Optional`, the old defensive defaults were load-bearing.

**Missing types**

If the SDK has other block types (e.g. `ImageBlock`, future variants), they'll silently be dropped here. That's probably fine for now, but a final `else: logger.debug("unhandled block type", block_type=type(block).__name__)` would make future debugging much easier.

**Tests**

PR body says "manual testing completed" but no automated test changes. Given this fixes a crash/regression (#161), a unit test with a mock `ThinkingBlock` and `TextBlock` would prevent it regressing. Worth a nudge.

**Bottom line:** Merge-worthy with the `ThinkingBlock` behavior either confirmed intentional or adjusted. The `isinstance` refactor is the right call.

PR #163, opened 2026-03-23
fix: preserve session on timeout instead of destroying it

gabmfranco-ds:
## Summary

- **Bug**: When Claude SDK times out during a session resume, the bot treats it as "session expired" and destroys the session. The next user message starts a fresh session with no conversation history, causing the bot to say "I don't have context about what we were doing."
- **Fix**: Catch `ClaudeTimeoutError` separately from other resume errors. Timeouts are transient — the Claude session is still valid on the CLI side. The session is preserved so auto-resume works on the next message.
- **Timeout bump**: Default timeout increased from 300s to 1200s (20 min) to better support long-running agentic operations.

## Root cause

In `facade.py`, the `except Exception` block (line 92) caught **all** errors during session resume — including timeouts — and ran the "stale session" recovery path: delete session → create new → retry without history. But `ClaudeTimeoutError ≠ session expired`. The session is likely still alive on Claude's side; it just took too long to respond.

## Changes

| File | Change |
|------|--------|
| `src/claude/facade.py` | Catch `ClaudeTimeoutError` before the generic `Exception` handler; preserve session and touch `last_used` to prevent expiration |
| `src/utils/constants.py` | `DEFAULT_CLAUDE_TIMEOUT_SECONDS`: 300 → 1200 |

## Test plan

- [ ] Send a message that triggers a long-running Claude operation
- [ ] Verify that after timeout, the next message resumes the same session (not a fresh one)
- [ ] Verify that a genuine session-expired error still creates a fresh session (existing behavior preserved)

### DIFF
--- a/src/claude/facade.py
+++ b/src/claude/facade.py
@@ -11,7 +11,6 @@ from typing import Any, Callable, Dict, List, Optional
 import structlog
 
 from ..config.settings import Settings
-from .exceptions import ClaudeTimeoutError
 from .sdk_integration import ClaudeResponse, ClaudeSDKManager, StreamUpdate
 from .session import SessionManager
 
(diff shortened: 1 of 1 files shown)


RichardAtCT:
Good fix for a real bug — silent session destruction on timeout is nasty. A few things worth addressing before merge:

---

**1. String matching is fragile (biggest concern)**

The PR title says "Catch `ClaudeTimeoutError` separately" but the implementation never actually imports or checks `ClaudeTimeoutError`. Instead it does substring matching on error messages — which is the opposite approach to what's described. If `claude-agent-sdk` exposes typed exceptions (and the PR body implies it does), use them:

```python
from claude_agent_sdk import ClaudeTimeoutError, ClaudeSessionExpiredError

try:
    ...
except ClaudeSessionExpiredError:
    # session is gone, rebuild
except ClaudeTimeoutError:
    # transient, preserve
except Exception:
    # unknown — preserve or reraise?
```

If typed exceptions don't exist yet, that's worth a `# TODO` and a comment explaining why string matching is the fallback. As written, this is brittle — a future SDK error message change silently reverts the bug.

---

**2. The `should_continue=False AND session_is_gone=True` gap**

When `should_continue` is `False` and `session_is_gone` is `True`, the code falls into the `else` branch: it touches `session.last_used`, logs "preserving session", and re-raises. That means a genuinely dead/expired session is preserved in storage and will cause the same expired-session error on the next message. Probably not intended.

The four combinations should be explicit:

| `should_continue` | `session_is_gone` | Current behaviour |
|---|---|---|
| True | True | rebuild session ✅ |
| True | False | preserve + raise ✅ |
| **False** | **True** | preserve + raise ⚠️ — dead session saved |
| False | False | preserve + raise ✅ |

Suggest adding a comment or separate branch for `not should_continue and session_is_gone`.

---

**3. `_SESSION_GONE_HINTS` defined inside the except block**

Hoist to module level with a type annotation (mypy strict requires it):

```python
_SESSION_GONE_HINTS: tuple[str, ...] = (
    "session not found",
    "invalid session",
    "session expired",
    "no such session",
)
```

---

**4. The 300→1200s timeout bump**

4× increase is a big jump. This will hold a Telegram handler open for up to 20 minutes. Questions:
- Is there user-facing feedback during the wait (typing indicator)?
- Does python-telegram-bot's handler have its own timeout that fires before 1200s?
- Would 600s cover real-world cases with less risk?

Warrants a comment explaining the rationale, and ideally a per-request configurable timeout.

---

**5. No tests**

No new tests for a regression fix. At minimum:
- A test that a simulated timeout does **not** call `session_manager.remove_session`
- A test that a "session not found" error **does** call `remove_session`

These are the exact two paths now split — they're the most important things to pin.

---

**Summary:** The intent is correct and the bug is real. But the implementation diverges from the described approach (typed exception vs. string matching), has a logic gap in the `should_continue=False/session_is_gone=True` case, and ships with no tests. Request changes on those three points before merging.

PR #165, opened 2026-03-26
feat: per-chat routing, group prefix trigger, and conversation history

morellid:
## Summary

- **Per-chat working directories**: new `PERSONAL_CHAT_ID`/`PERSONAL_CHAT_DIRECTORY` and `GROUP_CHAT_ID`/`GROUP_CHAT_DIRECTORY` settings map specific Telegram chat IDs to dedicated Claude sessions and working directories. Unrecognised chats fall back to `APPROVED_DIRECTORY`.
- **Session isolation per chat**: all session state (`claude_session_id`, `force_new_session`, etc.) moved from `context.user_data` (per-user) to `context.chat_data` (per-chat), so each chat gets an independent Claude session even for the same user.
- **Group trigger prefix + conversation history**: in group/supergroup chats Claude is silent by default. Only messages starting with `GROUP_TRIGGER_PREFIX` (default: `claude`) — or the slash form `/claude` — produce a response. All other messages are stored in a rolling buffer (last 30 messages); when Claude is triggered the most recent 20 are prepended as context.

## New files

- `src/bot/features/chat_routing.py` — `get_working_directory()` routing helper and `GroupChatBuffer` rolling history buffer

## Changed files

- `src/config/settings.py` — 5 new optional fields + validators
- `src/bot/orchestrator.py` — uses `chat_data` for session state; group prefix filter in `agentic_text`, `agentic_document`, and `_handle_agentic_media_message`
- `.env.example` — documents the new settings

## Test plan

- [ ] DM the bot → Claude uses `PERSONAL_CHAT_DIRECTORY`, session is independent from group
- [ ] Send a plain message in the group → no response, message stored in buffer
- [ ] Send `claude <request>` or `/claude <request>` in the group → Claude responds with recent conversation as context
- [ ] `/new` in DM resets only the DM session, group session unaffected
- [ ] No routing config set → behaviour identical to before (falls back to `APPROVED_DIRECTORY`, no group filtering)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -189,6 +189,8 @@ GROUP_CHAT_DIRECTORY=
 # Prefix word required to trigger Claude in group chats.
 # Messages NOT starting with this prefix are silently buffered for context
 # but do not produce a response. Supports both "claude <msg>" and "/claude <msg>".
+# Buffered history is stored in Telegram ``chat_data`` and only survives bot
+# restarts when a python-telegram-bot Persistence backend is configured.
 # Default: claude
 GROUP_TRIGGER_PREFIX=claude
 
--- a/.env.example
+++ b/.env.example
@@ -168,6 +168,30 @@ WHISPER_CPP_BINARY_PATH=
 # Named models look for ~/.cache/whisper-cpp/ggml-{name}.bin
 WHISPER_CPP_MODEL_PATH=base
 
+# === PER-CHAT ROUTING ===
+# Route specific Telegram chats to dedicated Claude sessions and working directories.
+# This allows, for example, a personal DM to work in your home directory while a
+# shared group chat works in a project-specific folder — each with its own session.
+
+# Telegram chat ID of your personal DM with the bot
+# Find your ID by messaging @userinfobot
+PERSONAL_CHAT_ID=
+
+# Working directory for the personal DM session
+PERSONAL_CHAT_DIRECTORY=
+
+# Telegram chat ID of a group chat (negative number, e.g. -1001234567890)
+GROUP_CHAT_ID=
+
+# Working directory for the group chat session
+GROUP_CHAT_DIRECTORY=
+
+# Prefix word required to trigger Claude in group chats.
+# Messages NOT starting with this prefix are silently buffered for context
+# but do not produce a response. Supports both "claude <msg>" and "/claude <msg>".
+# Default: claude
+GROUP_TRIGGER_PREFIX=claude
+
 # === PROJECT THREAD MODE ===
 # Enable strict routing by Telegram project topics
 ENABLE_PROJECT_THREADS=false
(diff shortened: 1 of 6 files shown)


RichardAtCT:
Good concept — per-chat routing and group silencing are genuinely useful features. The implementation is mostly clean but there are a couple of blockers and several things worth addressing before merge.

---

### 🔴 Blockers

**1. No type hints in `chat_routing.py`**

Violates the project's `mypy strict` + `disallow_untyped_defs=true` requirement. Every function needs full annotations:

```python
def append(buffer: list[dict[str, str]], sender_name: str, text: str) -> None: ...
def format_history(messages: list[dict[str, str]]) -> str: ...
def get_working_directory(chat_id: int, settings: Settings) -> str: ...
```

**2. No automated tests**

The group trigger logic has meaningful branching (exact match, prefix match, slash variant, stripping, history injection) and `get_working_directory` has three paths. All of these are pure functions — trivially testable with pytest. Manual-only is not acceptable for a feature this size.

---

### 🟡 Non-blocking issues

**3. Redundant slice guard**

```python
# Current
context_messages = history[-HISTORY_CONTEXT_SIZE:] if len(history) > HISTORY_CONTEXT_SIZE else history
# Simplified — slice handles the short-list case natively
context_messages = history[-HISTORY_CONTEXT_SIZE:]
```

**4. `getattr` fallback should be direct access**

`group_trigger_prefix` is now a defined settings field — use `self.settings.group_trigger_prefix` directly instead of `getattr(..., "claude")`.

**5. Telegram slash commands with `@botname` suffix**

In groups, Telegram often appends the bot username: `/claude@mybotname`. The current slash prefix logic won't match this form. Worth handling:

```python
slash_variants = (slash_prefix + " ", slash_prefix + "@")
```

**6. `chat_data` buffer persistence is implicit**

Buffer is stored in `chat_data` which only persists across restarts if a `Persistence` object is configured. Worth a note in `.env.example` or a startup warning when no persistence is configured.

**7. Prefer telegram constants**

```python
from telegram import Chat
if chat_type in (Chat.GROUP, Chat.SUPERGROUP):
```

**8. `GroupChatBuffer` is a namespace, not a class**

A class with only `@staticmethod` methods and no instance state is just a module with functions in a trench coat. Either give it real state (own the buffer) or make them module-level functions.

---

### ✅ What's good

- Moving session state from `user_data` → `chat_data` is the right call for group isolation
- `buffer[:-1]` to exclude the triggering message from history context is a nice touch
- Blank-string → `None` validators in settings are solid defensive config handling
- Directory existence validation at startup catches misconfiguration early

---

**Summary:** Fix the type hints and add at least basic unit tests for the pure functions in `chat_routing.py` and the trigger logic. The `@botname` slash issue is worth a fix too — it'll bite immediately in any real group deployment.

RichardAtCT:
Great design — the separation into chat_routing.py is clean, and the rolling buffer approach for group context is clever.

**Action needed**: We're merging a batch of 7 PRs right now that will change main significantly (especially orchestrator.py). Please rebase on main after the batch completes and verify tests pass. The routing logic itself looks good to merge once conflicts are resolved.

Will review again after rebase.

RichardAtCT:
Please can you fix the failing test and then we should be good to go!

PR #166, opened 2026-03-26
Fix: Explicitly set proxy for httpx client to prevent connection pool corruption

bibotai:
## Problem

When running behind a proxy (e.g., Clash, V2Ray), the bot's polling loop can get stuck even if `HTTP_PROXY`/`HTTPS_PROXY` environment variables are set. The httpx connection pool becomes corrupted after network interruptions, and the bot stops responding to messages while the process is still running.

## Root Cause

python-telegram-bot's `Application.builder()` does not automatically use the `HTTP_PROXY`/`HTTPS_PROXY` environment variables for its httpx client. The proxy needs to be explicitly set via `builder.proxy()`.

## Solution

Explicitly read proxy from environment variables and pass it to the Application builder:

```python
import os
proxy_url = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
if proxy_url:
    builder.proxy(proxy_url)
    logger.info("Proxy configured", proxy=proxy_url)
```

## Testing

- Bot now correctly uses proxy for Telegram API calls
- Polling loop remains stable after network interruptions
- Tested with Clash proxy on macOS

## Related Issues

This partially addresses issues where bots stop responding after running for some time behind proxies.

### DIFF
--- a/src/bot/core.py
+++ b/src/bot/core.py
@@ -64,6 +64,18 @@ class ClaudeCodeBot:
         builder.write_timeout(30)
         builder.pool_timeout(30)
 
+        # Explicitly set proxy from environment variables.
+        # This is necessary because python-telegram-bot's Application.builder()
+        # does not automatically use HTTP_PROXY/HTTPS_PROXY environment variables.
+        # Without this, the httpx connection pool can become corrupted when running
+        # behind a proxy, causing the bot to stop responding to messages.
+        import os
+
+        proxy_url = os.environ.get("HTTPS_PROXY") or os.environ.get("HTTP_PROXY")
+        if proxy_url:
+            builder.proxy(proxy_url)
+            logger.info("Proxy configured", proxy=proxy_url)
+
         self.app = builder.build()
 
         # Initialize feature registry


RichardAtCT:
**PR Review**
Reviewed head: `d837d3f723e4c29dc932532715820648787ed3b9`

**Summary**
- Explicitly passes proxy URL from env vars to `Application.builder()` to fix connection pool corruption when running behind a proxy (Clash, V2Ray, etc.)

**What looks good**
- Correct use of `builder.proxy()` — this is the right PTB API for this
- Sensible env var priority: `HTTPS_PROXY` before `HTTP_PROXY`
- Logs when proxy is active, silent when not — good operational behaviour

**Issues / questions**

1) **[Important]** `src/bot/core.py` — `import os` inside a method body is non-idiomatic; `os` is almost certainly already imported at the top of this file. Move the import there if not already present.

2) **[Important]** `src/bot/core.py` — Logging the full proxy URL (`proxy=proxy_url`) will expose credentials if the URL contains a username/password (e.g. `http://user:pass@host:port`). Redact before logging:
   ```python
   from urllib.parse import urlparse
   parsed = urlparse(proxy_url)
   safe = parsed._replace(password="***").geturl() if parsed.password else proxy_url
   logger.info("Proxy configured", proxy=safe)
   ```

3) **[Minor]** The PR doesn't address the *connection pool corruption* aspect — it only ensures the proxy is used. If the polling loop truly gets stuck after network interruptions, there may be a separate reconnect/retry issue worth a follow-up ticket.

**Test gaps**
- No test coverage. At minimum, a unit test that mocks `os.environ` with `HTTPS_PROXY` set and asserts `builder.proxy()` is called would provide a regression guard.

**Verdict**
⚠️ Merge after fixes — credential leak in logs is a real concern; `import os` placement is a quick cleanup.

PR #189, opened 2026-04-18
feat: add VERBOSE_LEVEL=3 to preserve progress message after response

tonylampada:
Currently the progress message (Claude's tool-call log + reasoning shown during execution at VERBOSE_LEVEL>=1) is deleted as soon as the final response is sent, so users lose the record of what tools were invoked and with what inputs.

Extend VERBOSE_LEVEL from range 0-2 to 0-3, where the new level 3 keeps the progress message visible after completion (Stop button stripped, text preserved as a read-only audit log). Levels 0-2 behave exactly as before.

Rationale: a single verbosity knob keeps configuration simple -- no cross-flag dependency between "how verbose" and "keep the log" to reason about.

Centralize the delete/keep decision in a new
MessageOrchestrator._finalize_progress_msg helper used by agentic_text, agentic_document, and the media handler. The /verbose command and Pydantic validator now accept 0|1|2|3. README, CLAUDE.md, and docs/tools.md updated.

Example output

CleanShot 2026-04-19 at 09 44 06@2x


tonylampada:
Disclaimer
This was done via claude code telegram

i did not look at the code

PR #192, opened 2026-04-24
feat: accept PDF uploads in agentic mode

timurvafin:
## Summary

Add `.pdf` to the allowed-extensions whitelist so the security middleware lets PDF files through. In `agentic_document`, branch early for PDFs: save the file to `<approved_directory>/.uploads/<timestamp>-<name>` and pass Claude an absolute path plus an instruction to use the Read tool. Claude reads PDFs natively via its built-in Read tool.

## Motivation

Users reported that sending a PDF (ticket, receipt, invoice) to the bot produced `🛡️ File Upload Blocked — File type not allowed: .pdf`. Adding only the whitelist entry wasn't enough — the downstream handler then tries to UTF-8 decode the bytes and fails with `Unsupported file format`. This PR makes both paths work for PDFs.

## Changes

- `src/security/validators.py`: add `.pdf` to `ALLOWED_EXTENSIONS`
- `src/bot/orchestrator.py`: new `_save_pdf_and_build_prompt` helper + early branch in `agentic_document` for PDFs
- `.gitignore`: ignore `.uploads/`
- Tests: whitelist coverage, `.pdf.exe` regression, full `agentic_document` PDF flow (tmp_path based)

## Known limitations

- Classic mode (`AGENTIC_MODE=false`) is intentionally untouched. PDFs there would still hit the UTF-8 decode branch. Can be addressed in a follow-up PR.
- `.uploads/` has no automatic cleanup. Manual housekeeping for now — a separate PR could add a `/new` hook or cron-based LRU.
- Only `.pdf` is added. `.docx`/`.xlsx` require different UX (Bash + converters) and are out of scope here.

## Test plan

- [x] `make test` — 532 passed
- [x] `black`, `isort`, `flake8` clean on changed files
- [x] Live smoke on a local bot: sent a PDF containing `SMOKE-TOKEN-42`, verified (a) no block message, (b) file saved to `.uploads/<timestamp>-ticket.pdf` with matching size, (c) Claude's response included the token via Read tool
- [x] Regression unit tests: `.exe` still blocked, files > 10 MB still rejected, `.pdf.exe` trap name still blocked

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -146,4 +146,5 @@ data/
 sessions/
 backups/
 uploads/
+.uploads/
 config/mcp.json
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -9,6 +9,7 @@ import asyncio
 import re
 import time
 from dataclasses import dataclass, field
+from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional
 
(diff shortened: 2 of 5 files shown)


timurvafin:
Superseded by #193, which implements the same capability through the proper FileHandler abstraction. #193:

- routes PDFs through the existing FileHandler (new `"document"` type) instead of a PDF-specific branch in `agentic_document`
- gives classic mode PDF support for free
- opens a clean extension point for future binary formats

Please review #193 instead.

PR #194, opened 2026-04-24
feat: include reply/quote context in Claude prompt

timurvafin:
## Summary

When a user replies to a message (especially with a **highlighted fragment** — Telegram's partial-quote feature, Bot API 7.0+), the bot was reading only `update.message.text` and dropping the quoted context. Claude then had to guess what the user was referring to.

This PR extracts the quoted fragment and prepends it to the user's new text as a markdown blockquote, so Claude sees both the fragment and the response in a single prompt.

## Before vs after

**Before:**
```
User (replying with quote on Claude's earlier message): "а каково его население?"
  → prompt sent to Claude: "а каково его население?"
  → Claude: "its what? please clarify"
```

**After:**
```
User (replying with quote on "Столица Франции — Париж."): "а каково его население?"
  → prompt sent to Claude:
    > Столица Франции — Париж.

    а каково его население?
  → Claude: "Население Парижа — около 2,1 млн человек…"
```

## Design

- **Priority:** `message.quote.text` (partial quote, Bot API 7.0+) wins over the full `reply_to_message.text`. If the user highlighted a fragment, that's what they mean.
- **Fallback:** If there's no partial quote, use the full replied-to message. If that message is media, use its `caption`.
- **Format:** markdown blockquote (`> line\n> line\n\ntext`), natural for Claude.
- **No changes** when there's no reply/quote — behaviour is identical.
- **Shared helper** in `src/bot/utils/quote_prompt.py` — both `agentic_text` and classic `handle_text` go through it, so classic mode gets the feature for free.

## Changes

- `src/bot/utils/quote_prompt.py` — new `build_user_prompt(message) -> str` helper with priority logic and markdown-blockquote rendering.
- `src/bot/orchestrator.py` — `agentic_text` uses `build_user_prompt(update.message)` instead of `update.message.text`.
- `src/bot/handlers/message.py` — `handle_text_message` (classic) same wiring; `.pdf` unaffected.
- `tests/unit/test_bot/test_quote_prompt.py` — 10 tests: plain/reply/partial-quote priority/multiline/media-caption/edge cases (no-text, no-reply, empty user text).

## Live verification

Reply via Telegram client with highlighted fragment of Claude's previous answer; bot's database prompt now includes the blockquote; Claude's reply is contextually correct.

## Test plan

- [x] `make test` — 540 passed (10 new)
- [x] `black`/`isort`/`flake8` clean
- [x] Live smoke: replied to a bot message with a follow-up question whose only referent was the quoted fragment. Claude resolved the reference correctly.

### DIFF
--- a/src/bot/handlers/message.py
+++ b/src/bot/handlers/message.py
@@ -25,6 +25,7 @@ from ..utils.image_extractor import (
     should_send_as_photo,
     validate_image_path,
 )
+from ..utils.quote_prompt import build_user_prompt
 
 logger = structlog.get_logger()
 
@@ -298,7 +299,9 @@ async def handle_text_message(
 ) -> None:
     """Handle regular text messages as Claude prompts."""
     user_id = update.effective_user.id
-    message_text = update.message.text
+    # Include reply/quote context so Claude sees the fragment the user is
+    # responding to, not just their new text.
+    message_text = build_user_prompt(update.message)
     settings: Settings = context.bot_data["settings"]
 
     # Get services
--- a/src/bot/orchestrator.py
+++ b/src/bot/orchestrator.py
@@ -39,6 +39,7 @@ from .utils.image_extractor import (
     should_send_as_photo,
     validate_image_path,
 )
+from .utils.quote_prompt import build_user_prompt
 
 logger = structlog.get_logger()
 
@@ -917,7 +918,9 @@ class MessageOrchestrator:
     ) -> None:
         """Direct Claude passthrough. Simple progress. No suggestions."""
         user_id = update.effective_user.id
-        message_text = update.message.text
+        # Include reply/quote context so Claude sees the fragment the user is
+        # responding to, not just their new text.
+        message_text = build_user_prompt(update.message)
 
         logger.info(
             "Agentic text message",
--- a/src/bot/utils/quote_prompt.py
+++ b/src/bot/utils/quote_prompt.py
@@ -1,0 +1,77 @@
+"""Build prompts from Telegram messages, including reply/quote context.
+
+When a user replies to a message (especially with a highlighted fragment —
+Telegram's partial-quote feature, Bot API 7.0+), the bot's default behaviour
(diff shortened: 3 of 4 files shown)


RichardAtCT (changes requested):
Nice change — small, well tested, and wired the same way in both modes. Tests and lint pass.

One thing before I merge. The quoted text goes into the prompt as a bare markdown blockquote, with nothing telling Claude where it came from (`src/bot/utils/quote_prompt.py:38`).

In a group, the replied-to message can come from anyone, not just an allowed user. Please add a short label, for example "Quoted message (reference only):", so Claude treats it as material to read rather than instructions to follow.

One line. Then I will merge.

FrundlesTian:
Hi @timurvafin — I prepared the small follow-up requested in the review:

- Prefix quoted context with `Quoted message (reference only):`.
- Update the existing quote-prompt tests for the labeled output, including the quote-only case.

The commit is based directly on the current PR HEAD (`7035eb7`) and contains no unrelated changes:

https://github.com/FrundlesTian/claude-code-telegram/commit/f16ef49431a67fba503731f7846d9c9346ba475e

It can be applied with:

```bash
git fetch https://github.com/FrundlesTian/claude-code-telegram.git fix/pr-194-quote-label
git cherry-pick f16ef49431a67fba503731f7846d9c9346ba475e
```

Verification: all 540 tests pass; Black, isort, flake8, and mypy are clean.

FrundlesTian:
Hi @RichardAtCT — quick review update: I prepared a single-commit patch for the outstanding request to label quoted content as reference-only context, including updates to all existing quote-prompt tests.

The patch is based directly on the current PR HEAD (`7035eb7`):
https://github.com/FrundlesTian/claude-code-telegram/commit/f16ef49431a67fba503731f7846d9c9346ba475e

All 540 tests pass, and Black, isort, flake8, and mypy are clean. The PR branch itself has not been updated yet; it is currently waiting for @timurvafin to cherry-pick the patch. Once applied, it should address the remaining review request.

PR #131, opened 2026-03-06
feat: passthrough unknown slash commands to Claude in agentic mode

haripatel07:
## Problem
In agentic mode, unknown slash commands (e.g. `/workflow`, `/skill`, `/add-memory`) 
were silently dropped because the `TEXT` handler excluded all `COMMAND` messages via 
`filters.TEXT & ~filters.COMMAND`.

## Fix
Remove `~filters.COMMAND` from the agentic text `MessageHandler` so unrecognised 
slash commands are forwarded to Claude as natural language.

Registered commands (`/start`, `/new`, `/status`, etc.) still take priority via 
higher-priority `CommandHandler`s and are completely unaffected.

Classic (non-agentic) mode retains the original filter — unknown commands are 
not forwarded there.

## Change
`src/bot/orchestrator.py` line ~321:
```python
# Before
filters.TEXT & ~filters.COMMAND

# After  
filters.TEXT  # unknown slash commands pass through to Claude in agentic mode
```

## Related Issue
Closes #129

## Type of Change
- [x] Bug fix / feature (restoring expected agentic behaviour)

## Testing
- [x] 491 passed, 0 failed
- [x] No regressions

Issue #176, opened 2026-04-07
Support configurable bot commands in agentic mode


sakhnenkoff:
## Problem

Related to #173.

In agentic mode, `get_bot_commands()` returns a hardcoded list of commands. Users who define custom slash commands (forwarded to Claude via `_handle_unknown_command`) have no way to make those commands appear in Telegram's `/` command menu without patching `orchestrator.py` directly.

The passthrough mechanism works great — unrecognized `/commands` are forwarded to Claude as natural language. But without menu visibility, users have to remember their custom commands or type them manually.

## Proposed Solution

Allow users to define additional bot commands via environment variable or config file. These would be added to `get_bot_commands()` for menu visibility only — they would NOT be added to `_known_commands`, so they continue to pass through to Claude via `_handle_unknown_command`.

Example config approach:
```
CUSTOM_BOT_COMMANDS=today:Morning briefing,closeday:End of day review,capture:Quick capture
```

Or via a YAML/JSON config section.

I have a simpler implementation that adds specific commands directly — happy to open a PR as a starting point or discuss the config approach.

## Environment

- claude-code-telegram v1.6.0
- Agentic mode enabled

xiaotonng:
Hardcoded command menus that don't reflect user-defined slash commands create a frustrating discoverability gap. In pikiclaw, custom workflows are first-class: project-level skills in `.pikiclaw/skills/` (compatible with `.claude/commands/*.md` format) are automatically surfaced in the Telegram `/skills` menu without any config patching. If you want to take the YAML/env-var config approach further in a more extensible direction, the skills system design in our repo would be worth a look — and contributions to extend it are very welcome at `npx pikiclaw@latest`. GitHub: https://github.com/xiaotonng/pikiclaw

Issue #173, opened 2026-04-06
Support custom bot commands in Telegram menu via config


sakhnenkoff:
Currently `get_bot_commands()` in the orchestrator hardcodes the list of commands that get registered with Telegram's `/` autocomplete menu via `set_my_commands`.

This means if you have project-level Claude Code slash commands (defined in `.claude/commands/`), they work fine when typed manually — the bot forwards them to Claude via `_handle_unknown_command` — but they don't show up in Telegram's command menu, so users don't know they exist.

Since `set_my_commands` runs on every startup, you can't work around this with BotFather either — the bot just overwrites whatever you set there.

**Proposal:** Support a config option (env var or config file) for additional bot commands to register alongside the built-in ones. Something like:

```
# .env
CUSTOM_BOT_COMMANDS=today:Morning briefing,closeday:End of day review,capture:Quick capture a thought
```

Or a JSON/YAML file:

```yaml
# custom_commands.yaml
- command: today
  description: Morning briefing
- command: closeday
  description: End of day review
```

These would just be appended to the existing command list in `get_bot_commands()`. The bot already handles unknown commands correctly by forwarding them to Claude, so this is purely a UI/discoverability improvement.

Happy to open a PR if you're open to it.

xiaotonng:
I'm building a similar tool (`pikiclaw`) that bridges Claude Code / Codex / Gemini to Telegram / Feishu / WeChat plus a localhost dashboard, where **this works out of the box — drop a file at `.pikiclaw/skills/<name>/SKILL.md` (or any `.claude/commands/*.md`) and the next session picks it up as a `/sk_<name>` slash command in Telegram's autocomplete, no env-var list to maintain and no config edits**. Under the hood the file system is the source of truth, descriptions are read from each skill's frontmatter, and the same `set_my_commands` call just merges custom skills with the built-ins. One reference shape.

`npx pikiclaw@latest`
https://github.com/xiaotonng/pikiclaw

PR #197, opened 2026-04-27
feat: discover Claude Code skills + Telegram command name normalization

americodias:
## Summary

Discovers Claude Code skills from the standard on-disk locations and surfaces them as Telegram bot commands so they appear in the command menu and can be invoked with `/<skill-name>`. Also handles the dash-to-underscore normalization Telegram requires, transparently rewriting back before forwarding to Claude.

## Why

Claude Code skills are first-class extensibility: drop a `SKILL.md` into `.claude/skills/<name>/`, with a YAML frontmatter `name:` and `description:`, and Claude can invoke it. The bot already passes through unknown slash commands to Claude (#131), which means skill commands like `/git-activity` *do* reach Claude — but:

1. They're invisible in Telegram's command menu (no autocomplete).
2. Telegram's Bot API only allows `[a-z0-9_]` in command names, so any skill named `git-activity` (the conventional dashed form) can't be a `BotCommand` at all. Users have to remember to type `/git_activity` and that doesn't match Claude's skill dispatcher, which uses the raw frontmatter name.

This PR fixes both: discovers skills from the same paths Claude Code itself uses, exposes them with normalized names in the menu, and rewrites the command back to the original form before passing to Claude.

## What

**`src/bot/features/skill_discovery.py`** (new, 172 LOC) — scans the standard skill locations:

```
{project_dir}/.claude/skills/<skill>/SKILL.md                    (project)
~/.claude/skills/<skill>/SKILL.md                                (user)
~/.claude/plugins/marketplaces/<m>/{plugins,external_plugins}/<p>/skills/<s>/SKILL.md
```

with project > user > plugin precedence. Parses YAML frontmatter for `name`, `description`, `argument-hint`. Skips skills declaring `user-invokable: false` (so non-user-facing skills like agent-only ones don't pollute the menu) and skips collisions with built-in commands (`start`, `new`, `status`, etc).

The `rewrite_skill_command(text, skills)` helper maps a leading `/<normalized>` back to `/<original_name>` for discovered skills — leaves non-command text, unknown commands, and already-original-form commands untouched.

**`src/bot/orchestrator.py`** (+30 LOC) — wires it in:

- `__init__`: scans skills once at startup
- `agentic_text`: rewrites the leading slash before forwarding to Claude
- `agentic_new`: re-scans on `/new` so newly-added skills appear without restarting
- `get_bot_commands`: appends discovered skills to the agentic command list

**`tests/unit/test_bot/test_skill_discovery.py`** (new, 225 LOC) — 20 unit tests covering the multi-path discovery, precedence, shadowing, normalization, command rewrite, and `user-invokable: false` filter.

## Compatibility

- Pure addition for projects without `.claude/skills/` directories — `discover_skills()` returns an empty dict, no behavior changes.
- Uses the existing `passthrough unknown slash commands` plumbing from #131 — slash commands that don't match a registered handler still flow through to Claude as before. The rewrite step is a no-op when the command isn't a discovered skill.
- No new dependencies. `PyYAML` is already present in `pyproject.toml`.

## Test plan

- [x] 20 unit tests passing (`pytest tests/unit/test_bot/test_skill_discovery.py -v`)
- [x] Tested live against a project with 51 project-level + 21 plugin-level skills — discovery surfaced all 72 in the Telegram command menu, dashed names rendered as `_` and round-tripped correctly through Claude
- [ ] Test in a project with no `.claude/skills/` (graceful no-op)
- [ ] Test plugin-only skills (no project skills) resolution

## Notes

The default `_BUILTIN_COMMANDS` list (skipped during discovery) is hardcoded to match the agentic-mode handlers (`start`, `new`, `status`, `verbose`, `repo`, `tts`, `restart`, `help`, `sync_threads`). If a user names a skill the same as a built-in, the built-in wins and a debug log notes the skip — this matches Claude Code's own behavior.

### DIFF
--- a/src/bot/features/skill_discovery.py
+++ b/src/bot/features/skill_discovery.py
@@ -1,0 +1,172 @@
+"""Discover Claude Code skills from project, user, and plugin locations.
+
+Scans SKILL.md frontmatter for name, description, and argument-hint.
+Returns a dict of skill name -> DiscoveredSkill. Project-agnostic -- works
+with any Claude Code project, following the standard on-disk skill layout:
+
+  {project_dir}/.claude/skills/<skill>/SKILL.md                        (project)
+  ~/.claude/skills/<skill>/SKILL.md                                    (user)
+  ~/.claude/plugins/marketplaces/<m>/plugins/<p>/skills/<s>/SKILL.md   (plugin)
+  ~/.claude/plugins/marketplaces/<m>/external_plugins/<p>/skills/<s>/SKILL.md
+
+Precedence: project > user > plugin. On collision the higher-precedence
+entry wins, and the shadowed one is logged at debug.
+"""
+
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Dict, Iterable, Optional, Tuple
+
+import structlog
+import yaml
+
+logger = structlog.get_logger(__name__)
+
+# Commands that must not be overridden by skills
+_BUILTIN_COMMANDS = frozenset({
+    "start", "new", "status", "verbose", "repo", "tts",
+    "help", "sync_threads", "restart",
+})
+
+
+@dataclass
+class DiscoveredSkill:
+    name: str                   # Telegram-safe form (lowercase, [a-z0-9_])
+    description: str
+    argument_hint: Optional[str] = None
+    original_name: str = ""     # raw `name:` from frontmatter (may contain dashes)
+    source: str = "project"     # "project" | "user" | "plugin"
+
+
+def _normalize(raw_name: str) -> str:
+    return raw_name.strip().lower().replace(" ", "_").replace("-", "_")
+
+
+def _iter_skill_files(project_dir: Path) -> Iterable[Tuple[Path, str]]:
+    """Yield (SKILL.md path, source) pairs in precedence order."""
+    project_dir = Path(project_dir)
+    # Project skills: <project>/.claude/skills/<skill>/SKILL.md
+    project_skills = project_dir / ".claude" / "skills"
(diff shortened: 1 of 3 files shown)


RichardAtCT (changes requested):
This is the right answer to #173 and #176 and I want it. Four things first.

**1. A long or numerous skill set stops the bot booting.** `get_bot_commands` appends every discovered skill with no cap. Telegram allows 100 commands and 32 characters per name, and `_normalize()` does not truncate. `set_my_commands()` runs uncaught in `src/bot/core.py:142`, so a `BadRequest` kills startup rather than degrading the menu.

Please truncate names to 32 characters, cap the list, and catch the error so a bad menu never blocks boot. Your own note says you discovered 72 skills — that is close to the limit.

**2. Discovery reads the operator personal home directory.** `skill_discovery.py:62` walks the real `~/.claude/skills` and `~/.claude/plugins`, with no way to inject a root or turn it off. On a shared bot, the commands strangers see depend on what the operator happens to have installed.

It also breaks an existing test: `test_agentic_bot_commands` expects 6 commands and gets 39 on a machine with skills installed. Please take the roots as a parameter so tests and CI are hermetic.

**3. Strip the `@botusername` suffix.** `rewrite_skill_command` does `head[1:].lower()` with no `@` handling. In a group Telegram sends `/git_activity@mybot`, so the rewrite misses. `_handle_unknown_command` already does this at `orchestrator.py:1550` — reuse that. This bot supports group mode, so it matters.

**4. Run `make format`.** `black --check` fails on both new files.

Item 1 is the one I care most about. The rest are quick.

PR #200, opened 2026-04-28
ci: align pipeline.yml to aphrollo-api/web standard

trickstertwo:
## Summary

Brings claude-code-telegram CI in line with the aphrollo-api / aphrollo-web pipeline standard (also used by lens, fetch, docflow, askdoc).

- **Move lint+test off ubuntu-latest → self-hosted.** Drops the GitHub-hosted minutes cost. Resolves the "re-runs lint+test on push-to-main after PR already passed" concern (still re-runs, but free now).
- **Add `vulncheck` job** — `poetry run pip-audit` against the lockfile. Python equivalent of govulncheck.
- **Add `secret-scan` job** — gitleaks v8.30.1 against full git history (`fetch-depth: 0`).
- **Pin `actions/checkout`** to v6.0.2 SHA, matching Go pipelines.
- **Extend `deploy.needs`** to gate on vulncheck + secret-scan.

Plus `.gitleaks.toml` allowlist for the curl-auth-header false positives in `tests/unit/test_orchestrator.py` (test fixtures) and `docs/setup.md` (example commands). Verified locally: gitleaks goes from 3 findings → 0.

## Test plan

- [ ] PR run: lint, test, vulncheck, secret-scan all green on self-hosted runner
- [ ] vulncheck (\`pip-audit\`) passes against current \`poetry.lock\` — if it surfaces real CVEs, fix in a follow-up PR rather than expanding allowlist
- [ ] Deploy job stays gated (\`if: workflow_dispatch || (push && main)\`) — should not run on this PR
- [ ] Merge to main triggers full pipeline including deploy

## Notes

- Did NOT touch \`release.yml\` (separate concern, runs on tag push only).
- \`release.yml\` still runs lint+test on \`ubuntu-latest\`. Open question for follow-up: align it too, or accept that release-on-tag is rare enough to keep using GitHub-hosted runners?


trickstertwo:
Wrong base repo — should target aphrollo/claude-code-telegram (the fork we run from). Reopening there.

Issue #208, opened 2026-05-30
Question: DeepSeek provider compatibility — similar to OpenRouter use case (#171)


phlips5437:
Hi @RichardAtCT, thanks for this great project.

I'm using Claude Code CLI with **DeepSeek** as the model provider (via Claude Code's built-in provider configuration), not Anthropic's API directly. Before deploying this bot, I'd like to understand the compatibility:

**My setup:**
- Claude Code configured with DeepSeek backend (deepseek-v4-pro)
- No `ANTHROPIC_API_KEY` set — authentication goes through Claude Code's own provider routing
- Plan to use `USE_SDK=false` (CLI subprocess mode) to avoid SDK-level API assumptions

**Questions:**

1. **CLI mode (`USE_SDK=false`):** Since this just shells out to `claude` CLI, does it fully work with any model provider that Claude Code supports? My understanding is yes — the bot just runs `claude -p "..."` and the provider routing is transparent. Can you confirm?

2. **SDK mode (`USE_SDK=true`):** I noticed #171 reports "No content to display" with OpenRouter. Does the SDK integration make assumptions about Anthropic-specific response formats that break with third-party providers?

3. **`ANTHROPIC_BASE_URL` (#143):** Is there any timeline for merging custom endpoint support? This would help users of OpenRouter, DeepSeek, and other compatible APIs.

4. **General direction:** Is third-party provider support something you'd consider in-scope for this project? Happy to help test/reproduce if needed.

**Context:** OpenRouter user reported the same class of issue in #171 (one-line fix in `sdk_integration.py:569`). DeepSeek likely hits the same code paths.

snalipleasegamble-source:
I was using claude with glm and used codex to set it up for me, took me a lot of time to convince codex because even he gave up during the setup, lol. At the end, it did work. Idk how but somehow

Issue #171, opened 2026-04-04
Bug: "No content to display" when using OpenRouter as backend


x7007x:
## Problem
When using OpenRouter as ANTHROPIC_BASE_URL, bot always returns "No content to display".

## Root Cause
ResultMessage.result returns '' (empty string) with OpenRouter, not None.

Line 569 in sdk_integration.py:
`if result_content is not None:` → True for empty string → returns blank content

## Fix (one line)
Change line 569 from:
if result_content is not None:

To:
if result_content:

## Tested
OpenRouter + qwen/qwen3.6-plus:free ✅

PR #143, opened 2026-03-10
feat: add anthropic_base_url setting for custom API endpoints

fdkgenie:
- Add anthropic_base_url field to Settings class with corresponding property
- Update ClaudeSDKManager to set ANTHROPIC_BASE_URL environment variable
- Add documentation to .env.example, README.md, and docs/configuration.md
- Enables flexible configuration of proxy/enterprise endpoints for Anthropic API

PR #210, opened 2026-07-30
feat: add MiniMax provider configuration

octo-patch:
Reason: provider-add - configure MiniMax through the existing Claude Agent SDK extension point while preserving the current SDK/session flow.

## Changes

- Add `CLAUDE_PROVIDER=minimax` and `MINIMAX_REGION` settings.
- Resolve the global `https://api.minimax.io/anthropic` and China `https://api.minimaxi.com/anthropic` endpoints automatically.
- Default MiniMax to `MiniMax-M3` and allow `MiniMax-M2.7` through `CLAUDE_MODEL`.
- Keep `ANTHROPIC_BASE_URL` as an explicit override and leave the SDK client/session flow unchanged.
- Document the provider, regions, endpoints, and model choices.

## Verification

- `pytest tests/unit/test_config.py` - 31 passed.
- Focused SDK endpoint/model tests - 8 passed.
- `black --check`, `isort --check-only`, `flake8`, and `mypy src/config/settings.py` - passed.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -63,6 +63,12 @@ USE_SDK=true
 # Get your API key from: https://console.anthropic.com/
 ANTHROPIC_API_KEY=
 
+# Custom base URL for the Anthropic API (optional)
+# Use this to point the SDK at a proxy/enterprise endpoint or an
+# Anthropic-compatible provider endpoint. Leave empty for the default endpoint.
+# Example: https://your-proxy.example.com/anthropic
+ANTHROPIC_BASE_URL=
+
 # Path to Claude CLI executable (optional - will auto-detect if not specified)
 # Example: /usr/local/bin/claude or ~/.nvm/versions/node/v20.19.2/bin/claude
 CLAUDE_CLI_PATH=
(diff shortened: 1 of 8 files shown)


octo-patch:
Updated this PR with MiniMax provider selection, global and China endpoint resolution, and MiniMax-M3/MiniMax-M2.7 model configuration. I ran 31 configuration tests, 8 focused SDK tests, black, isort, flake8, and mypy.

RichardAtCT:
Thanks for this. It is clean and well tested, but I am not merging a per-vendor block.

There are four open requests for the same thing: MiniMax here, OpenRouter in #171, DeepSeek in #208, and a generic base URL in #143. Adding one vendor at a time means a settings block per vendor forever.

I want one `ANTHROPIC_BASE_URL` plus model mapping that answers all four. That is the next thing I plan in this area.

If you want to pick that up instead, say so and I will point you at the shape. It also conflicts with main now, after #220 rewrote the surrounding block.

PR #217, opened 2026-09-11
feat: interactive Telegram approval for risky tool calls

### LINKED ISSUE #216: Feature: Interactive tool permission approval via Telegram (like CLI's approve/deny prompt)
## Is your feature request related to a problem?

The Telegram bot's tool permission checking (`can_use_tool` callback in `src/claude/sdk_integration.py`) is purely automatic: it validates file paths and bash command boundaries against static rules (`SecurityValidator`, `check_bash_directory_boundary`) and immediately returns `PermissionResultAllow()` or `PermissionResultDeny()`.

This differs from the Claude Code CLI experience, where the user is interactively prompted to approve or deny a risky tool call (e.g. running a bash command, editing a file) before it executes. In the Telegram bot today there is no way for a human to be put in the loop for a specific tool call — it's either allowed by the static rules or silently denied, with no per-call approval step.

For users running the bot with looser security settings (e.g. `DISABLE_SECURITY_PATTERNS=true`) or for higher-risk actions in general, this removes a safety net that the CLI provides natively.

## Describe the solution you'd like

Add an optional interactive approval mode where, before executing a tool call that matches configurable criteria (e.g. `Bash`, `Write`, `Edit`, or any tool not on a pre-approved allowlist), the bot:

1. Sends a Telegram message to the user showing the tool name and a summary of its input (command / file path / diff preview)
2. Presents inline keyboard buttons, e.g. `✅ Allow once`, `✅ Allow for session`, `❌ Deny`
3. Blocks the `can_use_tool` callback (with a timeout) until the user responds via callback_query, then returns the corresponding `PermissionResultAllow()` / `PermissionResultDeny()`
4. On timeout, defaults to deny (fail closed) and lets the user know

This should be configurable (e.g. `INTERACTIVE_TOOL_APPROVAL=true/false`, default off to preserve current behavior) so it doesn't break existing automated/headless usage.

## Describe alternatives you've considered

- Keep the current static allow/deny-only model and rely solely on `claude_allowed_tools`/`claude_disallowed_tools` config — simpler but offers no per-call human oversight for edge cases the static rules don't anticipate.
- Use the SDK's native `permission_mode` (e.g. `default`/`plan`) instead of a custom Telegram-specific approval UI — worth evaluating, but the SDK's built-in prompt is designed for a terminal, not Telegram's inline keyboard UX, so a custom callback + inline keyboard bridge is likely still needed.

## Additional context

Relevant code:
- `src/claude/sdk_integration.py:180-238` — `_make_can_use_tool_callback`, the async callback that would need to pause and wait for a Telegram response
- `src/config/settings.py` — no `permission_mode` or interactive-approval setting currently exposed
- `CLAUDE.md` security model section documents the current 5-layer static defense (auth → directory isolation → input validation → rate limiting → audit logging) with no interactive layer

I'd like to work on this — happy to open a PR.

FrundlesTian:
## Description

Adds an opt-in interactive approval step before Claude executes risky tool calls (default: `Bash`, `Write`, `Edit`) — similar to the Claude Code CLI's approve/deny prompt, but via a Telegram Allow/Deny inline keyboard.

When `INTERACTIVE_TOOL_APPROVAL=true`, the `can_use_tool` SDK callback additionally requests human approval (after the existing static path/bash-boundary checks pass) and blocks with a timeout; the outcome on timeout is configurable (deny by default, or allow).

## Related Issue

Closes #216

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

## How it works

- `ClaudeIntegration.run_command` / `_execute` / `ClaudeSDKManager.execute_command` now accept an optional `approval_callback`, threaded the same way `interrupt_event` already is for the Stop button.
- `MessageOrchestrator` builds this callback per-request (`_make_tool_approval_callback`), sending an Allow/Deny inline keyboard and tracking the pending prompt in a small dict keyed by a request id, resolved by a new `tapv:` callback_query handler modeled directly on the existing `stop:` handler.
- `StopAwareUpdateProcessor`'s priority-bypass list is extended to include `tapv:`, for the same reason the Stop button needs it: without it, a button click queues behind the very handler it's supposed to unblock, deadlocking.
- **Important**: the SDK's `can_use_tool` callback is only consulted for tools *not already* in `allowed_tools` — a listed tool is pre-approved by the CLI and never reaches the callback. Since `claude_allowed_tools` includes `Bash`/`Write`/`Edit` by default, gated tools are automatically excluded from the `allowed_tools` sent to the SDK while `INTERACTIVE_TOOL_APPROVAL` is enabled, or the prompt (and the pre-existing static checks) would silently never fire. This was found via live testing against a real bot — see testing notes below.
- New settings (all default to current/off behavior, so no behavior change unless explicitly enabled):
  - `INTERACTIVE_TOOL_APPROVAL` (default `false`)
  - `INTERACTIVE_TOOL_APPROVAL_TOOLS` (default `Bash,Write,Edit`)
  - `INTERACTIVE_TOOL_APPROVAL_TIMEOUT_SECONDS` (default `60`)
  - `INTERACTIVE_TOOL_APPROVAL_TIMEOUT_ACTION` (default `deny`, fail closed; can be set to `allow`)

Scoped to Allow/Deny only for this first PR (no "allow for the rest of the session" persistence) to keep the change reviewable — happy to follow up with that as a separate enhancement if there's interest.

## Testing
- [x] Tests added/updated
- [x] All tests pass (`make test` — 552 passed)
- [x] Manual testing completed — tested end-to-end against a real Telegram bot + live Claude session. All four paths verified: Allow (incl. a sandbox-escape retry with `dangerouslyDisableSandbox`, which correctly re-prompted), Deny, timeout→deny (default), timeout→allow (`INTERACTIVE_TOOL_APPROVAL_TIMEOUT_ACTION=allow`). Details in the PR comment below.

## Checklist
- [x] `make format` has been run (pre-commit hooks installed via `make dev`)
- [x] Self-review completed
- [x] Documentation updated (`CLAUDE.md`, `docs/configuration.md`, `.env.example`)
- [x] No breaking changes (new settings default to current behavior)


RichardAtCT (changes requested):
Thanks for this. Finding the `autoAllowBashIfSandboxed` bypass was the valuable part — #220 built on it, and it is now fixed on main.

Two changes before merge:

1. The prompt shows too little to decide on. `_summarize_tool_input()` returns the filename only for Write and Edit, and cuts Bash at 80 characters. Please show the full path, and a longer command window. It was written for verbose logging, not for an approval decision.

2. Register the pending approval before you send the keyboard (`src/bot/orchestrator.py:1767`). A fast click finds no entry today, so it waits out the full timeout.

Both are small. The rest looks right — static checks run before the prompt, the priority prefix avoids the deadlock, and you check the approver identity.

Note the allowlist fix here is behind the feature flag, so it does not close #219 on its own. That is handled on main already. No action needed from you.

